PR #2, opened 2015-12-01
GBM prediction code

davharris:
This code loops through the species and uses GBM to make predictions about the (Poisson-distributed) abundances.  It would take a while with 8 cores, so we should probably use Hipergator.

### DIFF
--- a/munge/train-test.R
+++ b/munge/train-test.R
@@ -1,0 +1,75 @@
+full_bbs = na.omit(readRDS("data/munged.Rds"))
+library(fastICA)
+library(magrittr)
+library(dplyr)
+library(beepr)
+
+focal_year = 1990
+
+set.seed(1)
+
+# Train on the training data through focal_year
+train_bbs = full_bbs[full_bbs$in_train & full_bbs$year <= focal_year, ]
+
+make_rescaler = function(x){
+  scaled_x = scale(x)
+  
+  # Half the variance in X should come from temp and half from precip,
+  # but 3/4 of the columns are about temperature.
+  temp_columns = grep("^t", colnames(scaled_x))
+  scaled_x[ , temp_columns] = scaled_x[ , temp_columns] / sqrt(3)
+  
+  ica = fastICA(scaled_x, 10)
+  
+  function(new_x){
+    # Z-transform the new variables using the old means and sds
+    scaled_new_x = sapply(
+      1:ncol(new_x), 
+      function(i){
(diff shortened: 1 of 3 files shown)


munge/train-test.R
```diff
@@ -0,0 +1,75 @@
...
+  scaled_x = scale(x)
+  
+  # Half the variance in X should come from temp and half from precip,
+  # but 3/4 of the columns are about temperature.
+  temp_columns = grep("^t", colnames(scaled_x))
+  scaled_x[ , temp_columns] = scaled_x[ , temp_columns] / sqrt(3)
+  
+  ica = fastICA(scaled_x, 10)
+  
+  function(new_x){
```
ethanwhite: This must be something I don't understand about R yet. How does this nested function end up getting executed?

ethanwhite:
Looks great overall.

The one main thing is my confusion as to how the execution of the unnamed function happens. Assuming this is basically a lambda function of some kind, then for something this length I think it's probably easier to understand as it's own separate (non-nested) named function, but since I don't really follow how it's being executed this may be something you have to do to get things to work.

The only other thing that I noticed is that I think a lot of the data management code could be made easier to read using `dplyr`, but that may just be a personal preference. E.g.,

```
train_climate = select(train_bbs, matches("ppt|tmin|tmean|tmax"))
train_rescaler = make_rescaler(train_climate)
```
or even

```
train_rescaler = make_rescaler(select(train_bbs, matches("ppt|tmin|tmean|tmax"))
```

instead of

```
climate_colnames = grep("ppt|tmin|tmean|tmax", colnames(train_bbs), value = TRUE)
train_rescaler = make_rescaler(train_bbs[ , climate_colnames])
```

ethanwhite:
hey @davharris - just wanted to see if you might have the chance to finish the recommended changes so that we can get this merged in. I'm also happy to go ahead and merge this one as is if you don't want to make any changes.

davharris:
Ack.  When did this merge happen? I think someone needs to take the command line away from me whenever I get sick.

Sorry!

davharris:
Oh no.  I set weecology/bbs-forecasting as my `upstream` remote *and* as my `origin` remote. So when I tried to push to my own master branch, I ended up pushing here.  Fixing now...

Really sorry!

ethanwhite:
No worries. Just go ahead and submit a new PR with any changes you want to make and reference this one so that it's easy to go back and look at the conversation.

PR #3, opened 2016-01-18
Remove use of git-lfs for handling database

ethanwhite:
git-lfs was difficult to work with due to the no public forks issue:
https://medium.com/@megastep/github-s-large-file-storage-is-no-panacea-for-open-source-quite-the-opposite-12c0e16a9a91

This removes the intermediate database from version control. For the moment
the user will need to generate it themselves (which in most cases would be the
case outside of the lab anyway). For in-house purposes I've put a copy of the T-drive.


ethanwhite:
@davharris or @sdtaylor, please review this when you get a chance and merge if this sounds reasonable to you. Thanks.

sdtaylor:
:+1:

ethanwhite:
Thanks @sdtaylor. Go ahead and push the big green merge button.

PR #16, opened 2016-02-11
Add bioclim

sdtaylor:
This script only requires the file .data/bbs_data.csv, which is loaded inside forecasting-bbs-core.R , and the appropriate packages. 

If you source this script and run the command `bbs_bioclim=get_bioclim_data()`, it will do all the legwork of downloading prism data, extracting values, and calculating bioclim variables. It will also store bioclim and prism data in .data/bbsforecasting.sqlite to make things quicker in future calls.

### DIFF
--- a/get_prism_data.R
+++ b/get_prism_data.R
@@ -4,39 +4,110 @@ library(prism)
 library(raster)
 library(dplyr)
 library(tidyr)
+library(stringr)
 options(prism.path = "./data/prismdata")
 
-months <- c(1:12)
-clim_vars <- c("ppt", "tmin", "tmean", "tmax")
-for (month in months){
-  for (clim_var in clim_vars){
-    get_prism_monthlys(type=clim_var, year = 1981:2014, month = month, keepZip=F)
+yearsToUse=1981:2014
+
+#######################################################
+#Downloads the raw prism rasters into the folder specified above. 
+downloadPrism=function(){
+  months <- c(1:12)
+  clim_vars <- c("ppt", "tmin", "tmean", "tmax")
+  for (month in months){
+    for (clim_var in clim_vars){
+      get_prism_monthlys(type=clim_var, year = yearsToUse, month = month, keepZip=F)
+    }
   }
 }
 
-datadirs = dir(datapath)
-for (datadir in datadirs) {
-  bil_file = paste(datadir, '.bil', sep = "")
-  bil_file_path = file.path(datapath, datadir, bil_file)
-  sql_file = paste(datadir, ".sql", sep = "")
-  sql_file_path = file.path(datapath, datadir, sql_file)
-  system(paste("raster2pgsql -s 4326", bil_file_path, datadir, ">", sql_file_path))
-  system(paste("psql -d bbsforecasting -f", sql_file_path))
+
+#########################################################
+#Pretty postgres isn't used anywhere after adding the raw prism to it,
+#so i'll just comment this out for now. 
+#datadirs = dir(datapath)
+#for (datadir in datadirs) {
+#  bil_file = paste(datadir, '.bil', sep = "")
+#  bil_file_path = file.path(datapath, datadir, bil_file)
+#  sql_file = paste(datadir, ".sql", sep = "")
+#  sql_file_path = file.path(datapath, datadir, sql_file)
+#  system(paste("raster2pgsql -s 4326", bil_file_path, datadir, ">", sql_file_path))
+#  system(paste("psql -d bbsforecasting -f", sql_file_path))
+#}
+
+########################################################
+#Takes output of ls_prism_data() and makes sure all files
+#within a year range are there. 
+check_If_Prism_Files_Present=function(prismLS, years){
+  #Extract out all the variable names and dates
+  prismLS = prismLS %>%
+    rowwise() %>%
+    mutate(var=strsplit(files, '_')[[1]][2], yearMonth=strsplit(files, '_')[[1]][5]) %>%
+    mutate(year=as.integer(substr(yearMonth,1,4)), month=as.integer(substr(yearMonth, 5,6)), present=1) %>%
+    select(-files, -yearMonth)
+  
+  #Setup a list of what should be there
+  toCheck=expand.grid(year=years, month=1:12, var=c('ppt','tmax','tmean','tmin'))
+  
+  #Left join on the toCheck DF puts present=1 wherever that var/year/month combo was in the prismLS DF.
+  #As long as all of present==1 in toCheck, then all prism rasters are accounted for.
+  toCheck=toCheck %>%
+    left_join(prismLS, by=c('year','month','var'))
+  
+  if(sum(toCheck$present, na.rm=TRUE)==nrow(toCheck)){
+    return(TRUE)
+  } else {
+    
+    return(FALSE)
+  }
 }
 
-bbs_data <- read.csv("data/bbs_data.csv")
-locations <- unique(select(bbs_data, site_id, long, lat))
-coordinates(locations) <- c("long", "lat")
-#prism_files <- grep("_[0-9]{4}[0]",ls_prism_data()[,1],value=T)
-prism_stacked <- prism_stack(ls_prism_data())
-extracted <- raster::extract(prism_stacked, locations)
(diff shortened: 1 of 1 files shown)


get_prism_data.R
```diff
@@ -4,39 +4,232 @@ library(prism)
...
+#Years of prism data to 
+yearsToUse=1981:2014
+#Load the DB. create = TRUE does not seem 
+
+sqliteDBFile='./data/bbsforecasting.sqlite'
+if(file.exists(sqliteDBFile)){
+  database <- src_sqlite(sqliteDBFile, create = FALSE)
+} else {
+  database <- src_sqlite(sqliteDBFile, create = TRUE)
+}
```
ethanwhite: I think we can just use `database <- src_sqlite(sqliteDBFile, create = TRUE)` with none of the conditional logic. The documentation is unclear on this point, but a quick test suggests that `create` only comes into play if the file doesn't exist already (and this would be a pretty common approach to this sort of API).

get_prism_data.R
```diff
@@ -4,39 +4,232 @@ library(prism)
...
+  #As long as all of present==1 in the toCheck DF, then all prism rasters are accounted for.
+  toCheck=toCheck %>%
+    left_join(prismLS, by=c('year','month','var'))
+  
+  if(sum(toCheck$present, na.rm=TRUE)==nrow(toCheck)){
+    return(TRUE)
+  } else {
+    return(FALSE)
+  }
+}
```
ethanwhite: It seems like it would be cleaner to use `identical` to directly compare `prismLS` and `toCheck`. This would save both the merge and the summing and therefore the need to explain how all of this is working, and so could replace lines 66-75 with `return(identical(prismLS, toCheck))`

ethanwhite:
At the start of `get_prism_data()` `try` is used to check for the existence of the `prism_bbs_data`, i.e., to figure out if the database has been loaded or created. I think the logic could be made a little more straightforward here by using `prism_bbs_data %in% src_tbls(database)`.

get_prism_data.R
```diff
@@ -4,39 +4,232 @@ library(prism)
...
+    stop('Im stopping because the sqlite query for prism_bbs data returned neither a data frame nor an error in get_prism_data()')
+  }
+  
+}
+
+
+#################################################################
+#From raw prism monthly values calculate all the bioclim variables.
+#You should not call this directly to load bioclim vars. Instead call
+#get_bioclim_data(), which will 1st try to load the data from the sdqlite
```
ethanwhite: "sdqlite" -> "sqlite"

get_prism_data.R
```diff
@@ -4,39 +4,232 @@ library(prism)
...
+#db before processing it all from scratch.
```
ethanwhite: This function documentation should be directly attached to the function and the helper function gets it own separate docs (which it basically already has).

get_prism_data.R
```diff
@@ -4,39 +4,232 @@ library(prism)
...
-datadirs = dir(datapath)
-for (datadir in datadirs) {
-  bil_file = paste(datadir, '.bil', sep = "")
-  bil_file_path = file.path(datapath, datadir, bil_file)
-  sql_file = paste(datadir, ".sql", sep = "")
-  sql_file_path = file.path(datapath, datadir, sql_file)
-  system(paste("raster2pgsql -s 4326", bil_file_path, datadir, ">", sql_file_path))
-  system(paste("psql -d bbsforecasting -f", sql_file_path))
+process_bioclim_data=function(){
+  #Get the prism data. It's pulled from the sqlite DB or downloaded as needed.
```
ethanwhite: Comments like this are generally considered unnecessary because they just restate the information that you get by reading the code. If folks need to know where the PRISM data is coming from they can read `get_prism_data()`.

ethanwhite:
We generally want to try to avoid mixed naming conventions (e.g., pot_hole and camelCase for variables) in the same project. As a general style guide we tend to use [Hadley Wickham](http://r-pkgs.had.co.nz/style.html).

get_prism_data.R
```diff
@@ -4,39 +4,232 @@ library(prism)
...
-database <- src_sqlite("./data/bbsforecasting.sqlite", create = TRUE)
-mydata <- copy_to(database, prism_bbs_data, temporary = FALSE,
-                  indexes = list(c("site_id", "year", "month")))
+
+####################################################################
+#Load the bioclim variables from sqlite. if they aren't available,
+#load the prism data, process bioclim, load it in sqlite, then return 
+#bioclim as requested.
+
+#Possibly put a check here to make sure all years are accounted for?
```
ethanwhite: I generally recommend starting comments like this with `TODO: `. This lets us quickly grep the code base for things that are marked in this way and lets us make sure that we've taken care of all of them before submitting. Alternatively they can be added as issues in the github repo. Likewise, if something is a fix, rather than a to do, it's common to use `FIXME: ` in an equivalent way.

get_prism_data.R
```diff
@@ -4,39 +4,232 @@ library(prism)
...
+####################################################################
+
+get_bioclim_data=function(){
+  #Query the bioclim data and check to see if it throws an error. 
+  bioclim_bbs_data=try(collect(tbl(database, sql('SELECT * from bioclim_bbs_data'))))
```
ethanwhite: Similar comment to before about explicitly checking to see if the table exists rather than catching an error and interpreting the type.

get_prism_data.R
```diff
@@ -4,39 +4,232 @@ library(prism)
...
+  if(typeof(bioclim_bbs_data)=='list'){
+    #If no error, return the data. 
+    return(as.data.frame(bioclim_bbs_data))
+  } else if(class(bioclim_bbs_data)=='try-error') { 
+    #If erros on access the database, assume it's because the data just isn't there.
+    #So load it from scratch, store it for future access, and return the newly process bioclim data
+    print("bioclim data table not found, processing from scratch. ")
+    bioclim_bbs_data=process_bioclim_data()
+    
+    x=copy_to(database, bioclim_bbs_data, temporary = FALSE,
```
ethanwhite: Does this require assignment? Seems like you can probably just drop the `x=`.

ethanwhite:
This looks awesome @sdtaylor! Nice work.

Because it's in such great shape already I've made a bunch of comments related to style and making things easier to read. This is partly to help improve this particular code, but also just to start having a more general conversation about style and readability since your code is already so excellent.

Take what's useful and ping me when you're happy with your updates and we'll get this merged.

get_prism_data.R
sdtaylor: identical! awesome, didn't know that existed.

get_prism_data.R
sdtaylor: I would rather keep the conditional in case `create` changes in the future and, for example, throws an error if `create = TRUE` and the file already exists.

ethanwhite: OK.

ethanwhite: I've also opened a PR to see if they'll clarify this intended behavior for future use:
https://github.com/hadley/dplyr/pull/1664

sdtaylor:
I implemented all your other suggests. 
-simplified checking if tables already exists by using `table_name %in% src_tbls(database)`
-organized the comments better and fixed some typos
-changed all camelCase to pot_hole to match the rest of the project. 

Still working on using identical() in checking for all the prism rasters being present. It's very picky about characters vs factors. 
all.equal() does the same thing but is less strict, so I may try that.

ethanwhite:
That all looks great.

> It's very picky about characters vs factors. 

Can you just set `stringsAsFactors = FALSE` in `expand.grid`.

sdtaylor:
setting `stringsAsFactors = FALSE` does not work either with `identical`. But it does make `as.equal()` do what I'm trying to do.

sdtaylor:
Last suggestion implemented. 
-Simplified check for all the prism rasters by using all.equal()

ethanwhite:
Looks great!

get_prism_data.R
ethanwhite: FYI, this PR has been merged now, so we can count on `src_sqlite` to still read existing DBs even when `create = TRUE` from now on.

PR #18, opened 2016-03-02
Add a check for no PRISM files being present

ethanwhite:
check_if_prism_files_present needs to check if there are no PRISM files
present at all to avoid erroring if the directory is empty.

### DIFF
--- a/get_prism_data.R
+++ b/get_prism_data.R
@@ -54,6 +54,7 @@ download_prism=function(){
 #########################################################
 check_if_prism_files_present=function(prism_ls, years){
   #Extract out all the variable names and dates
+  if (length(prism_ls$files) == 0){return(FALSE)}
   prism_ls = prism_ls %>%
     rowwise() %>%
     mutate(var=strsplit(files, '_')[[1]][2], yearMonth=strsplit(files, '_')[[1]][5]) %>%


get_prism_data.R
```diff
@@ -54,6 +54,7 @@ download_prism=function(){
 #########################################################
 check_if_prism_files_present=function(prism_ls, years){
   #Extract out all the variable names and dates
+  if (length(prism_ls$files) == 0){return(FALSE)}
```
sdtaylor: can you add a comment like this?

#Check if prism directory is empty and no files are listed
if (length(prism_ls$files) == 0){return(FALSE)}
#If there are prism files present, extract out all the variable names and dates from their names
prism_ls = prism_ls %>%
.....

ethanwhite: This seems like [documenting mechanics instead of design and purpose](http://journals.plos.org/plosbiology/article?id=10.1371/journal.pbio.1001745#s8). If you don't what is happening is clear from the way the code is written I'd be happy to rewrite the code. E.g., we could use an if/else here instead of the if/return if that would be clearer, or an an explicit `num_files` variable.

sdtaylor: It's clear. But only because I know what all the variables are and where they came from. Someone looking thru this for the 1st time might have to do some unnecessary digging to figure out why a length check is needed and how that could happen.

PR #23, opened 2016-03-08
Add gimms

sdtaylor:
Functions to get GIMMS ndvi for all bbs sites for all years available.

-Downloads the raw GIMMS data if necessary.
-extracts all bbs routes for all years.
-filters out low quality values using the GIMMS QA flag. 
-inserts -99 for missing values.
-stores the data in the sqlite database for quick retrieval.

Outside the script you just need to call get_bbs_gimms_ndvi() to get everything. Which will take an hour or two to download and extract the 1st time around.


get_ndvi_data.R
```diff
@@ -1,39 +1,152 @@
-# Get NDVI data from MODIS
+#Functions for retrieving and processing NDVI data for BBS route locations. Currently only GIMMS is suported, with years 1981-2013 available.
```
ethanwhite: "suported" -> "supported"

get_ndvi_data.R
```diff
@@ -1,39 +1,152 @@
...
+#
+#GIMMS you just need to call get_bbs_gimms_ndvi() to retreive a dataframe of (site_id, year, month, ndvi).
```
ethanwhite: "retreive" -> "retrieve"

ethanwhite:
Is there a reason related to this data to use `-99` for null values (i.e., it's the standard for this type of data)? In general we try to avoid using numeric nulls (see Section 5 of [http://doi.org/10.4033/iee.2013.6b.6.f](http://doi.org/10.4033/iee.2013.6b.6.f) from some discussion). I haven't dealt with pushing nulls from R into an SQL database, but in Python it would properly convert them into SQL `NULL` values.

get_ndvi_data.R
```diff
@@ -1,39 +1,152 @@
...
+#This assumes we want all gimms files that are available. It queries for files
+#that are available for download and checks to see which, if any, are missing.
+#Returns TRUE if all files that are available are already present, and list of 
+#files to download if any are missing.
+##################################################
+check_if_gimms_files_present=function(){
+  available_files_download_path=sort(gimms::updateInventory())
+  available_files_name=basename(available_files_download_path)
+  
+  files_present=sort(list.files(gimms_folder))
```
ethanwhite: Is this sort needed elsewhere?

get_ndvi_data.R
```diff
@@ -1,39 +1,152 @@
...
+  #hdr files are created from some of the gimms processing that we don't want to
+  #use here.
+  files_present=files_present[!grepl('hdr', files_present)]
+  
+  if(length(files_present)>0){
+    to_download=available_files_download_path[! available_files_name == files_present]
+  } else {
+    to_download=available_files_download_path
+  }
```
ethanwhite: Is there a need for the conditional here? If there is nothing in `available_files_download_path` then line 38 should return `NULL` which has length == 0.

get_ndvi_data.R
```diff
@@ -1,39 +1,152 @@
...
+  
+  if(length(to_download)==0){
+    return(TRUE)
+  } else {
+    return(to_download)
+  }
```
ethanwhite: What about just returning `to_download` here regardless and then checking for `NULL` instead of `TRUE` when the function is called?

get_ndvi_data.R
```diff
@@ -1,39 +1,152 @@
...
+#Get the GIMMS AVHRR ndvi bi-monthly time series for every bbs site.
+#Pulling from the sqlite DB or extracting it from raw gimms data
+#################################################
+get_bbs_gimms_ndvi = function(){
+  if('gimms_ndvi_bbs_data' %in% src_tbls(database)){
+    return(collect(tbl(database, sql('SELECT * from gimms_ndvi_bbs_data'))))
+  } else {
+    print('Gimms NDVI bbs data not found, processing from scratch')
+    
+    file_status=check_if_gimms_files_present()
```
ethanwhite: Switching to just returning `to_download` above will let this variable name be more understandable, e.g., `files_to_download`.

get_ndvi_data.R
```diff
@@ -1,39 +1,152 @@
...
+    if(!isTRUE(file_status)){
```
ethanwhite: And then... `if(!is.null(files_to_download)` or maybe clearer `if(length(files_to_download) > 0)`

ethanwhite:
Looks great overall @sdtaylor! I've made a few minor suggestions. Ping me when you think it's ready to go in. Thanks!

get_ndvi_data.R
sdtaylor: Order did matter for doing this `available_files_name==files_present`. But just using == is actually not doing what I thought it was doing. It should be `available_files_name %in% files_present`. And using %in% order doesn't matter.

get_ndvi_data.R
sdtaylor: if files_present has 0 length then `available_files_download_path[! available_files_name == files_present]` gives a null value, when really it needs to return all files to download. But it's not needed after changing the `==` to `%in%`.

get_ndvi_data.R
sdtaylor: yep

sdtaylor:
-99 is my personal preference over R writing NA to csv's. But I can see why it can cause issues later on.
I checked passing NA's to sqlite and back and it looks like they are preserved, so I'll just stick to that.

ethanwhite:
Looks good. Thanks!

> -99 is my personal preference over R writing NA to csv's

Ah yes, I totally understand that desire. You can change these values when writing to csv using `na=`. I find just leaving them blank is the most robust way for them to be identified as nulls across different kinds of software, so `na = ""`.

PR #26, opened 2016-04-04
Add script for downloading and adding elevation data to database

ethanwhite:


ethanwhite:
I just updated this PR to check the database to see if the data is already there before proceeding.

sdtaylor:
should get_elev_data() be called download_elev_data() instead? Since it doesn't actually return the data.

ethanwhite:
That's a good point. In fact, for consistency do you think it would be better to rename that function and then add a new `get_elev_data()` function that actually queries the database the returns the values? Is that basically what you've done in the other functions.

sdtaylor:
yep, thats exactly what I do in the other scripts. get_xxx_data() downloads the data and puts it in the db if needed, but returns the values either way.

ethanwhite:
Thanks. Just updated the PR to match that behavior and some of your syntax choice.

Issue #8, opened 2016-02-08
Add summed SDM models to analysis


ethanwhite:
@davharris started some work on this in #2, but it is waiting on some fixes.

@sdtaylor is doing a lot of this modeling for population level work already, so we could also integrate that code at some point if it's sufficiently modularized to let us use it easily.

PR #30, opened 2016-04-19
Move data extraction/processing code into core

ethanwhite:
Also cleans up an old comment and reruns the notebook following the move.

### DIFF
--- a/forecast-bbs-core.R
+++ b/forecast-bbs-core.R
@@ -52,7 +52,6 @@ get_bbs_data <- function(){
       install_dataset('bbs')
     }
 
-    #FIXME: This query doesn't currently deal with poorly sampled species (e.g., nocturnal)
     bbs_query = "SELECT (counts.statenum * 1000) + counts.route AS site_id, routes.lati as lat,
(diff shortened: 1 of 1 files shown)


ethanwhite:
@davharris this commit gives you access to the data you need for #8 in `get_pop_ts_env_data()`. If you have time to give it a review then we can get it in an you can use that function to get the raw data for the SDMs.

davharris:
Mostly looks good to me.  Are there any guarantees that `get_pop_ts_env_data` and `get_richness_ts_env_data` will return the same set of routes?

ethanwhite:
Based on the code I think they have to. I played around with passing the output of `get_pop_ts_env_data` to `get_richness_ts_env_data` and working from there, but decided that, at least at a first glance, it would probably be more difficult to read.

davharris:
Okay.  The other option would be to de-duplicate some of the operations like filtering out NAs and filtering by year so that they couldn't get out of sync.  But if these operations are only used twice, then maybe that's overkill.

If you decide that would be overkill, I don't have any objections to merging.

ethanwhite:
Good call. What do you think about this instead.

PR #31, opened 2016-04-20
BBS changed column names for latitude and longitude

davharris:
Unless I screwed something up, I think I now have all the data and can actually start working on this :-)


ethanwhite:
Awesome. Thanks @davharris!

PR #32, opened 2016-04-20
specify dplyr namespace

davharris:
Not sure what's going on with the merge commit, but I think we can just squash it when we merge.


ethanwhite:
Thanks @davharris. That kind of merge commit typically results when you update a local branch to match upstream after you've already made changes. I squashed it out as suggested.

Issue #35, opened 2016-04-21
oddball species


davharris:
How do we want to handle oddball "species" (e.g. unidentified species, hybrid species & subspecies)? They're usually pretty identifiable in the data using regexes, but it's not obvious to me how best to handle them.  

My proposal: 

* For subspecies, it might be best to lump them in with the rest of their species.
* For hybrids and unidentified species, we might have to drop them.  

On a related note, did we make a decision somewhere about whether to remove super-rare species from the data set?

ethanwhite:
Your proposal sounds good to me.

Do we need to drop the super-rare's for the species distribution modeling?

davharris:
We’d have to treat them specially in some way, either by fitting a different model with many fewer degrees of freedom or by dropping them altogether, I think.  Dropping them is simplest from my perspective.

ethanwhite:
I have no objections.

davharris:
So, I just had a revelation: rare species don't contribute very much to richness.

For example, there are 72 species that showed up 50 or fewer times in the training data; removing them would only change the average richness from 57.26 to 57.19.  

So (as long as we only care about richness) it shouldn't matter what we do with them. I'll just leave them in.

ethanwhite:
Sounds good. There are some papers we can cite for this as well if we need them.

PR #36, opened 2016-04-21
Drop partially-identified taxa (see #35)

davharris:


forecast-bbs-core.R
```diff
@@ -28,18 +28,36 @@ install_dataset <- function(dataset){
...
+#' (e.g. "sp.") or were considered hybrids (e.g. "A x B").
 #'
 #' @param df dataframe containing an species_id column
 #'
 #' @return dataframe, filtered version of initial dataframe
 filter_species <- function(df){
-  bbs_data_filtered <- df %>%
+  
+  con <- dbConnect(RPostgres::Postgres(), dbname = 'postgres', 
+                   host = "localhost", password = "p", user = "postgres")
```
ethanwhite: I'd like to avoid explicitly passing credential here if possible. Are you having trouble getting it to work without Line 40?

ethanwhite:
I just add #38, so if the easiest way around this credential issue is to go ahead and make the move here that would be fine.

forecast-bbs-core.R
davharris: Oops, yeah.  Good catch. I finally got RPostgres to recognize my host (with PGHOST) and password (with . pgpass), but it still defaults to using "davidharris" as my username even though:

* I have `export PGUSER=postgres` in my .bash_profile
* When I run `psql` from the command line, it respects PGUSER

Anyway, I'll push a new commit without the credentials.

ethanwhite:
Thanks @davharris.

PR #40, opened 2016-04-22
Subspecies

davharris:
This combines the subspecies into their species-level category, like we discussed in #35.  It's not the nicest code I've ever written, but I'm not sure if there's a better way.

(Note that this PR is based off the branch in #36)


davharris:
I think this branch is ready for code review. Combining subspecies takes 3 minutes on my laptop, but it's probably not worth it to find a faster way, since we just need to run it once-ish.

ethanwhite:
Thanks @davharris. Since you branched this off of #36 and Github isn't very smart in that context (as you can see the earlier commits are still in this PR even though they're already merged) can you close this and open a new PR from the same branch. Thanks.

Issue #38, opened 2016-04-22
Move all database work to either SQLite or PostgreSQL


ethanwhite:
We currently have one (and maybe soon 2) calls to a postgres database. Since we're already using SQLite to store and pass around environmental data we should go ahead and add the BBS data to that database as well and extract it from there.

davharris:
Oh, I was all excited to use PostGIS for faster buffer cutting with the NLCD data now that I've basically gotten all of this nonsense working.  But running raster::extract overnight is probably just as good, so I have no objections.

ethanwhite:
Yeah, it's a tradeoff with ease of reproducibility. I think we should pick either Postgres or SQLite. SQLite will be easier for others since it doesn't require installing anything extra, but I'm OK with having an opinionated dependency here as well.

Also to consider when we're making this decision: @sdtaylor has just finished creating a more efficient version of extract for use with buffers that is apparently quite a bit faster.

davharris:
Whoa, nice @sdtaylor

ethanwhite:
I've updated the title to reflect a broader conversation about this. Thoughts on which DBMS we'd like to use throughout?

Major tradeoff: more power/speed with Postgres, but requires users to install and get PostgreSQL working in a way that the code can easily access the database (which as @davharris has experienced can be a bit of work).

sdtaylor:
My extract with buffer improvement is only  faster when you have many rasters of the same data. Like prism or bioclim. For a single raster it shouldn't be any better than just extract(). NLCD has 3 or 4 version though, so it would help some there.

davharris:
If we don’t need PostGIS for this project and the data sets don’t get any bigger, I don’t see any upside to using psql.

ethanwhite:
> My extract with buffer improvement is only faster when you have many rasters of the same data. Like prism or bioclim. For a single raster it shouldn't be any better than just extract(). NLCD has 3 or 4 version though, so it would help some there. 

Ah, got it. Thanks for the clarification @sdtaylor.

> If we don’t need PostGIS for this project and the data sets don’t get any bigger, I don’t see any upside to using psql.

OK, if @sdtaylor is on board with this then let's proceed with SQLite and swap it out in the future if we end up needing more power.

sdtaylor:
@ethanwhite have you done anything with this? I have to play around with this part of the code anyway in #61 so I thought I would switch things to sqlite while I'm at it.

ethanwhite:
[...]

PR #42, opened 2016-04-25
combine subspecies

davharris:
like #40, but with cleaner history

### DIFF
--- a/forecast-bbs-core.R
+++ b/forecast-bbs-core.R
@@ -28,7 +28,7 @@ install_dataset <- function(dataset){
 #'
 #' Removes waterbirds, shorebirds, owls, kingfishers, knightjars,
 #' dippers. These species are poorly sampled due to their aquatic or
-#' noctural nature. Also removes taxa that were either partially unidentified 
+#' noctural nature. Also removes taxa that were either partially unidentified
 #' (e.g. "sp.") or were considered hybrids (e.g. "A x B").
 #'
 #' @param df dataframe containing an species_id column
@@ -36,11 +36,11 @@ install_dataset <- function(dataset){
 #' @return dataframe, filtered version of initial dataframe
 filter_species <- function(df){
   species_table = get_species_data()
-  
+
   is_unidentified = function(names) {
     grepl("/|unid\\.|sp\\.| or |hybrid| X | x ", names)
   }
-  
+
   valid_taxa = species_table %>%
     filter(!is_unidentified(english_common_name)) %>%
     filter(!is_unidentified(spanish_common_name)) %>%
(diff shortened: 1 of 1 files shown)


ethanwhite:
@sdtaylor - can you give this a review when you get a few minutes.

sdtaylor:
does the package RPostgres work for both of you? The only one I have via CRAN is RPostgreSQL, which looks like it's exactly the same thing but with a different name.

davharris:
I just used the one that was used in `filter_species`.  It works fine for me, except it doesn’t seem to respect my `PGUSER` variable.

sdtaylor:
working around it for now. It shouldn't be an issue once everything is switched to sqlite in #38

ethanwhite:
https://github.com/rstats-db/RPostgres

forecast-bbs-core.R
```diff
@@ -35,28 +35,78 @@ install_dataset <- function(dataset){
...
+  # Add the abundances for subspecies that co-occurred at the same site in the
+  # same year. Takes about 3 minutes on my laptop.
+  combine = function(x){
+    x$abundance = sum(x$abundance)
+    x[1, ]
+  }
+  fixed_subspecies = df %>% 
+    filter(species_id %in% new_subspecies_ids) %>%
+    group_by(site_id, year, species_id) %>%
+    do(combine(.)) %>%
```
sdtaylor: This part takes 5-6 minutes on my laptop. But replacing `do(combine(.))` with `summarize(abundance = sum(abundance))` takes only a few seconds and gives the same answer. What's reason for the former?

davharris: Um.  Don't think there was actually a good reason, except that `summarize()` drops columns. But we could add those back in easily.  I'll do that now.

Thanks for pointing this out.

davharris:
@sdtaylor would you mind reviewing this version? Thanks for the dplyr::summarize suggestion.

forecast-bbs-core.R
```diff
@@ -28,35 +28,75 @@ install_dataset <- function(dataset){
...
-    filter(species_id < 3900 | species_id > 3910) %>%
-    filter(species_id < 4160 | species_id > 4210) %>%
-    filter(species_id != 7010) %>%
-    filter(species_id %in% identified_taxa)
+    filter(species_id %in% new_subspecies_ids) %>%
+    group_by(site_id, year, species_id) %>%
+    summarize(abundance = sum(abundance)) %>%
+    left_join(dplyr::select(df, -abundance),
+              by = c("site_id", "year", "species_id"))
+}
```
sdtaylor: without the bind_rows() call this is now returning only those subspecies that needed adjusting. I think this is what you want. it works for me and preserves the lat long columns. 

'''
   fixed_subspecies = df %>% 
    filter(species_id %in% new_subspecies_ids) %>%
    group_by(site_id, year, species_id, lat, long) %>%
    summarize(abundance = sum(abundance)) %>%
    ungroup()
  
  # bind the rows of the combined subspecies with the rows of everyone else
 df %>%
    filter(!(species_id %in% new_subspecies_ids)) %>%
    bind_rows(fixed_subspecies)
'''

davharris: Yikes, good catch.  Couldn't I also just omit `filter(species_id %in% identified_taxa)`? That's what I had intended to do.

Thanks again for the helpful comments on this.

davharris: Wait, nevermind.  Rereading....

davharris: Sorry for the triple comment.  I meant to say, "omit `filter(species_id %in% new_subspecies_ids)`"  That way, we'd never collapse down to just the new species.  Now that we're using summarize, which is fast, I don't see a reason to split out the subspecies anymore.

sdtaylor: ok that makes more sense. But now that I run it, the total rows in df isn't being reduced like it should be from aggregating species. It looks like the left join is adding things back in somehow. I think we can just get rid of the join and put  lat, long in the `group_by()` part to retain them. lat longs are unique to each site_id, so it's not adding "extra groups". 

df %>%
    group_by(site_id, year, species_id, lat, long) %>%
    summarize(abundance = sum(abundance)) %>%
    ungroup()

davharris:
Thanks @sdtaylor! I changed those lines like you suggested.  Ready to merge?

sdtaylor:
Yep, looks good.

PR #43, opened 2016-04-25
packaging/namespace

### LINKED ISSUE #39: Stop the namespace madness
I see at least four options for dealing with all the namespaces clobbering each other

1) Leave all the `library` calls and add `::` where needed

2) Use `::` everywhere

3) Use explicit namespaces using R's package infrastructure

4) Use [the `import` package](https://github.com/smbache/import)

I'm leaning toward 3.  Thoughts @ethanwhite @sdtaylor?

davharris:
This should fix #39 

some miscellaneous notes:

* Very little of this has been tested, so I added an "in progress" label for now.
* I didn't notice any cases where we actually used two functions by the same name, so this namespacing approach should solve the whole issue.
* R doesn't want hyphens in package names, so I used BBSforecasting. I could change it to something else (like bbs-forecasting) if that would be preferable
* R has strong opinions about how to handle the `data` folder. These opinions make more sense in a CRAN context than in our context; we should probably rename it.


davharris:
@ethanwhite when you have a moment, could you confirm that your notebooks run on your machine as of 37a4626? 

Thanks!

ethanwhite:
Looks like it's working as long as both the SQLite database and `bbs_data.csv` are in the data folder.

If not the 2nd code cell of the notebook gives me (I think due to the lack of `bbs_data.csv`:

```
Error in .local(res, n, ...): could not find function "columnToRownames"

Error in .local(res, n, ...): could not find function "columnToRownames"
```

But, I'd be surprised if everything was working without those files in place before, so I'd say :+1: for 37a4626 at the moment.

davharris:
Cool, thanks.

After googling around, it looks like maybe your version of DBI might not match your version of the database driver? [DBI deprecated and removed columnToRownames recently](https://github.com/rstats-db/DBI/issues/91)

When you have a moment, could you try getting the most recent versions of DBI, RSQLite, and RPostgres to confirm that it still works without `bbs_data.csv`

ethanwhite:
Thanks. Looks like that may have done it, but I won't know for sure before morning as it's busy doing what looks like a full download and install of all of the data.

ethanwhite:
The overnight run worked great. Pretty much a clean clone and then run it as long as the modules are sufficiently up to date. :+1: 

Is this ready to merge on your end?

davharris:
Yep. Thanks. 

Dave

davharris:
Ugh.  Merge conflict. Working on it now...

davharris:
I believe it's ready to merge again, whenever you're ready, @ethanwhite.

ethanwhite:
Should I merge a major reorganization days before giving a new talk with unfinished analyses... seems like a bad idea... oh well, here goes :fingers_crossed:

Issue #44, opened 2016-04-25
shear maps of SDM predictions


davharris:

davharris:
Three species with different-looking distributions

sdtaylor:
is this along the same lines as #8?

davharris:
It's for Ethan's talk. He wanted maps that looked like the one I posted in #45.  We should be able to close this once that branch gets merged, I think.

davharris:
(#8 is also covered by #45)

davharris:
fixed in #77

PR #45, opened 2016-04-25
Sdm

davharris:
In progress, but hopefully ready soon.  Also includes a function for shear maps (#44).

image


davharris:
@sdtaylor @ethanwhite this is ready for code review when one of you has time.

ethanwhite:
hey @davharris - what's the status of getting this up and running?

ethanwhite:
@sdtaylor - if you have some time would you mind giving this a review. Thanks.

sdtaylor:
Will do.

sdtaylor:
So if I run sdm.R inside rstudio I get this error when get_env_data() is called. 

`Error in get_elev_data() : could not find function "getData" `

Which looks like the `getData()` from the raster package is not being loaded correctly from the `importFrom` statement. 

But when I run the script from the command line with `Rscript sdm.R` this error comes up

```
Error in UseMethod("db_list_tables") : 
  no applicable method for 'db_list_tables' applied to an object of class "SQLiteConnection"
Calls: %>% ... match -> src_tbls -> src_tbls.src_sql -> db_list_tables
Execution halted
```
Which comes from one of the database calls in in `get_env_data()` but I'm not sure which. Any ideas?

sdm.R
```diff
@@ -0,0 +1,147 @@
...
+x = tidyr::expand(occurrences, site_id, year) %>%
+  inner_join(env, by = c("site_id", "year")) %>%
+  arrange(year, site_id)
+
+locations = distinct(x, site_id, lat, long)
+
+y = distinct(occurrences, year, species_id, site_id, .keep_all = TRUE) %>%
+  inner_join(select(x, year, site_id), c("year", "site_id")) %>%
+  mutate(presence = abundance > 0) %>%
+  select(-abundance, lat, long, start_time, month, day) %>%
```
sdtaylor: Having an issue here. `start_time` and `month` aren't found when doing this select statement. Could it be from different bbs_data.csv files or databases? I'm using the `./data/bbs_data.csv` file thats on the T drive.

sdtaylor:
Are SDM's going to be another model type to compare against ARIMA, spatial ndvi, etc.? If so, this could use a wrapper to return forecasts in the same format as the others.

ethanwhite:
> Are SDM's going to be another model type to compare against ARIMA, spatial ndvi, etc.?

Yes, that's the idea.

sdm.R
```diff
@@ -0,0 +1,147 @@
...
+n_clusters = 100
+n_folds = 10
+stopifnot(n_clusters %% n_folds == 0)
+
+# Load data ---------------------------------------------------------------
+occurrences = get_bbs_data(start_yr = 1982, end_yr = 2013, min_num_yrs = 25)
+
+env = get_env_data() %>%
+  filter_ts(start_yr, end_yr, min_num_yrs) %>%
+  inner_join(distinct(occurrences, site_id, lat, long, year),
```
sdtaylor: This `inner_join(distinct())` puts the `occurrence` and `species_id` columns in the `env` data frame, which ends up getting used in the models. That definitely doesn't look intentional unless your going for some sort of pseudo-joint sdm. =)

ethanwhite:
@davharris - just wanted to check in on the status of this PR.

sdm.R
```diff
@@ -0,0 +1,147 @@
...
+
+cv_train_rows = function(i){
+  i != fold_ids
+}
+cv_test_rows = function(i){
+  i == fold_ids
+}
+# fit gbm -----------------------------------------------------------------
+
+# Function to cross-validate a GBM model with different numbers of trees, then
```
sdtaylor: In my stuff I estimate the optimum number of trees with `gbm.perf()`. I think it might be faster than doing cv manually.

davharris:
replaced by #76

PR #46, opened 2016-04-28
Add dynamic regression modeling to time-series forecasts

ethanwhite:
Adds a new time-series regression model (linear regression with ARIMA errors) to the suite of time-series models.

### DIFF
--- a/R/forecast-bbs-core.R
+++ b/R/forecast-bbs-core.R
@@ -328,13 +328,16 @@ cleanup_multi_ts_forecasts <- function(ts_forecast_df, groups){
 #'       forecast, since when not hindcasting these may be different
 #' @importFrom forecast auto.arima forecast meanf naive
 #' @export
-get_ts_forecasts <- function(grouped_tsdata, timecol, responsecol, lag = 1){
+get_ts_forecasts <- function(grouped_tsdata, timecol, responsecol, exogcol, lag = 1){
+  get_train_data <- function(df, colname) df[[colname]][1:(nrow(df) - lag)]
+  get_test_data <- function(df, colname) df[[colname]][(nrow(df) - lag + 1):nrow(df)]
   do(grouped_tsdata,
-     timeperiod = .[[timecol]][(length(.[[responsecol]]) - lag + 1):length(.[[responsecol]])],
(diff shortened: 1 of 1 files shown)


davharris:
It's a bit hard for me to evaluate this with all the repeated double brackets. Could `.[[responsecol]][1:(length(.[[responsecol]]) - lag)]` be pulled out as a function?

ethanwhite:
I agree, but I'm not sure how to make that work with `do`. Can I pass `.` to a function?

davharris:
Yeah, you're already passing `.` to functions in your code.  It looks to me like you have two repeated patterns, which I've called f and g:

```
f = function(x, colname) x[[colname]][1:(nrow(x) - lag)]
g = function(x, colname) x[[colname]][(nrow(x) - lag + 1):nrow(x)]
```

I haven't tested it, so it's possible that I've made a typo somewhere, but I think you could do something like this: 

```
get_ts_forecasts <- function(grouped_tsdata, timecol, responsecol, exogcol, lag = 1){
  do(grouped_tsdata,
     timeperiod = g(., timecol),
     cast_naive = naive(f(., responsecol), lag),
     cast_avg = meanf(f(., responsecol), lag),
     cast_arima = forecast(auto.arima(f(., responsecol), seasonal = FALSE), h = lag),
     cast_exog_arima = forecast(auto.arima(f(., responsecol), xreg = f(., exogcol), seasonal = FALSE), xreg = g(., exogcol)),
     test_set = g(., exogcol)
  )
}
```

If `f and `g` had reasonable names (like "first chunk" and "last chunk" or something), I think it would improve the readability over the double brackets.

ethanwhite:
That makes sense. Thanks. Testing an improved version now.

ethanwhite:
OK, take a look at this one.

davharris:
The changes to `get_ts_forecasts` make sense to me.  I can't view the notebook diff, though.

ethanwhite:
Yeah, ignore the notebook diff, they still aren't very helpful at the moment. If you're happy with `get_ts_forecasts` go ahead and merge. Thanks for the helpful review!

PR #47, opened 2016-04-28
collect whole ndvi series

davharris:
The dev version of dplyr apparently only gets the first hundred thousand rows by default, and throws a warning.  This fixes it.

### DIFF
--- a/R/get_ndvi_data.R
+++ b/R/get_ndvi_data.R
@@ -118,7 +118,7 @@ get_bbs_gimms_ndvi = function(
   dir.create(gimms_folder, showWarnings = FALSE, recursive = TRUE)
 
   if('gimms_ndvi_bbs_data' %in% src_tbls(database)){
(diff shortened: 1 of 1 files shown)


ethanwhite:
For some reason I can't easily find the docs on `collect()` this mornign. Is that optional argument present in the current release as well?

davharris:
Looks like it’s not very well documented.  If I understand this diff correctly, the `n` argument won’t affect anything in older versions of dplyr but is necessary in the newest versions.

davharris:
Right.  Keep forgetting that github strips links. when I reply in email.

Here's the diff: https://github.com/hadley/dplyr/commit/52934e70d66fce5f91647832c0b0a43b12d3bce6#diff-7ce7e496e1ebf6a668326daf0c482dcbL591

ethanwhite:
OK, thanks. Seems to be working OK for me with a toy example. I forgot that R ignores extra arguments that it doesn't need, so that makes sense.

PR #49, opened 2016-04-28
Temporal data

davharris:
The main function here is `get_temporal_data`.  It reports, for each route/year combination, how many minutes after sunrise the route started (with negative numbers meaning it started before sunrise).

It also reports julian day ("yday"), which could be useful for seasonal effects.


R/forecast-bbs-core.R
```diff
@@ -72,8 +72,13 @@ combine_subspecies = function(df){
...
 
   df %>%
-    group_by(site_id, year, species_id, lat, long) %>%
+    group_by(site_id, year, species_id) %>%
     summarize(abundance = sum(abundance)) %>%
+    ungroup() %>%
+    left_join(
+      distinct(dplyr::select(ungroup(df), -abundance, -species_id)),
+      by = c("site_id", "year")
+    ) %>%
```
ethanwhite: Can you walk me through this set of changes starting on line 74. The previous version made sense to me (after updating the names re-sum the abundances to deal with cases where multiple sub-species were present at a site), but I don't fully follow what's going on now. Maybe also add a comment at the start of this pipe to explain what is happening.

R/forecast-bbs-core.R
```diff
@@ -113,15 +119,17 @@ get_bbs_data <- function(start_yr, end_yr, min_num_yrs){
                          JOIN bbs.routes
                            ON bbs.counts.statenum=bbs.routes.statenum
                            AND bbs.counts.route=bbs.routes.route
-                         WHERE bbs.weather.runtype=1 AND bbs.weather.rpid=101;"
+                         WHERE bbs.weather.runtype=1 AND bbs.weather.rpid=101
+                           AND bbs.weather.statenum!=3 AND bbs.weather.countrynum=840;"
```
ethanwhite: Do we want these filters to be explicit or do we just want the lack of available environmental data to take care of them? My first instinct is to let the availability of environmental data take care of the filtering so that if we go back and at point and add environmental data outside of the contiguous US we automatically start folding that ecological data in as well.

ethanwhite:
Looks great overall. I made a couple of minor in-line comments.

Nice work checking for and finding the issues with the 35031 & 35039 sites!

ethanwhite:
Also, looks like we have some merge conflicts now. Can you try rebasing against master? Thanks.

ethanwhite:
@davharris - just checking in on the status of this PR.

davharris:
replaced by #71

PR #59, opened 2016-07-18
Add maps of richness and environmental variables

ethanwhite:
To do this without making the notebook too sluggish the plots are saved and then
loaded as pngs in markdown. This change was also make for the observed-predicted
plot of the forecasts made by different models at different lags.


ethanwhite:
@davharris - this should help with the checks for data munging errors you mentioned. At a first glance these maps all look about right to me.

PR #60, opened 2016-08-09
WIP: Ar1 state space model

davharris:
Still left to do:

* clean up the R code for data import (e.g. make sure it's consistent with other time series models, remove vestigial data transformations from earlier versions of the model)
* make sure output format is consistent with other time series models
* add a prior on `y`? I'm getting [fairly bad mixing](https://twitter.com/davidjayharris/status/762999526437289984) for some of the time/place combinations when no observations were made
* review the stan code with @ethanwhite and/or @sdtaylor to make sure I'm not doing something completely stupid


sdtaylor:
can you outline the model and likelihood function? that would make it easier to look at the stan code.

davharris:
Sure. The likelihood is basically just line 40 and line 43. The rest is mostly either boilerplate or just slicing the data so the likelihood code points to the right place.

***Line 40 (33 in the older version):***
    y[(start+1):stop] ~ student_t(nu, beta * y[start:(stop-1)], sigma);

This line describes the "true" richness value in a given site/year combination (i.e. the richness you'd get if you averaged over a large number of hypothetical transects at that time/place).  This value is t-distributed with `nu` degrees of freedom, location at `beta * previous_y` and scale of `sigma`.  It's vectorized, so the `+1`'s and `-1`s are there to make sure that the `y` values on the left side match up with the `y` values from one time step earlier on the right.

This line gets called once per site in every MCMC iteration. For each site, `start` should point to the first year of data for the site, and `stop` should point to the last year of data for the site.  

------------

***Line 43 (36 in the older version):***
    observed ~ normal(y[index], nugget);

For the subset of `y` values that correspond to a site/year combination where there were observations, the observed value is normally distributed with mean equal to `y` and standard deviation equal to `nugget`. (the word "nugget" is sometimes used to describe the part of the distribution that doesn't get carried over through space or time)

----

The code in the `generated quantities` block should describe exactly the same model, except without the vectorization.  Lines 66 and 67 in this section should correspond to lines 40 and 43 above.

davharris:
(note that I've made some edits to the comment above, so it would probably be better to read it online rather than in your inbox)

stan-models/AR1.R
```diff
@@ -0,0 +1,158 @@
+library(dplyr)
+library(tidyr)
+library(rstan)
+devtools::load_all()
+
+start_yr <- 1982
+end_yr <- 2013
+min_num_yrs <- 25
+last_training_year <- 2003
+richness_w_env <- get_richness_ts_env_data(start_yr, end_yr, min_num_yrs) %>%
```
sdtaylor: Not sure how it affects the model yet, but `get_richness_ts_env_data()` is not enforcing `min_num_years`, there are only ~1000 sites that have >25 years of data.

davharris: Hmm.  This is the same code that's being used in `forecasting-bbs-R.ipynb`, I think.  Want to make a separate Issue to discuss it?

sdtaylor: Ok.

ethanwhite:
Hey @davharris - sorry it took so long for me to look at this. I saw @sdtaylor's comment, figured he'd be chatting with you about it and archived the email. You're always welcome to ping me if I haven't looked at something in a couple of weeks and you're still looking for feedback.

I've read through this and don't see anything obviously problematic, but I also don't feel like I managed to follow it through step by step. As a general rule it's pretty hard to meaningfully review large blocks of code like `AR1.R` without it being split up into smaller documented functions. The smaller functions make it easier to keep track of everything and also, as a result, allow reviews of bigger chunks of code to be done in pieces. So, I can review a function, do a meeting, review another function, etc.

The other big thing that would be useful here is integrating with the existing code base. I don't think that necessarily has to happen immediately, but IME it's often easier to get that done if you do it while you're working on the addition as opposed to as an afterthought once it's already merged. That said, I'm happy to merge this if you'd rather come back and do the integration later.

davharris:
Thanks Ethan.  Sorry I should have put the "in progress" label sooner.  

I've rewritten some of the Stan code to make it faster & more interpretable, and will make the changes you suggested to the R code in terms of pulling out functions and integrating them into the existing code base.

davharris:
By the way, the model's story isn't changing much as I tweak it, although it is becoming a bit clearer.  

Basically, the mean-reverting random walk part of the model has most sites eventually returning to relatively similar means (+/- 2 species or so), and that the differences among sites have more to do with where they happen to be on their random walks over the last few decades than with their underlying averages (i.e. the random walks move super slowly and each site could eventually end up looking very different from how it looks today). There's a lot of uncertainty in the posterior distribution about how important the underlying differences among sites are, and it could be more like +/-1 or +/-4 than the +/-2 I listed above.

Meanwhile, differences among observers contribute about +/- 5 species to the observed richness values, with much less uncertainty.

ethanwhite:
> Thanks Ethan. Sorry I should have put the "in progress" label sooner. 

Ah, yeah, no worries. I should have been able to infer that from the text but was moving kind of fast that evening. A good standard way to communicate this clearly in all formats is to add WIP: to the beginning of the title. This stands for Work In Progress and shows up in email notifications as well as on GitHub. When you think it's in good shape for a full review you then remove the WIP from the title and add a comment saying it's ready for a full review.

Glad to hear that the models story is settling in. It will be interesting to chat more about the interpretation while looking at the output.

davharris:
By the way, it looks (unsurprisingly) like this model is wrong: sites don't just happen to be at different points in their random walks: a site's average `ndvi_sum` explains half of the variance in the latent richness estimates (i.e. the estimate of how many species an average birder would be expected to find).

davharris:
I think the Stan code is pretty close to where I want it now.  It takes an arbitrary number of environmental predictors & predicts both the richness that the median observer would expect to see and a version that includes observer biases. For observers that aren't found in the training set, their bias is unknown and so the posterior distribution is larger.

The predicting all happens in the "generated quantities" section of the Stan code because it's faster (no MCMC needed for the predicted values) and because I learned the hard way that trying to make all the pointers line up correctly when I mix training and test data in the earlier code blocks creates a huge mess.

I still want to clean up the R code like we talked about, and finish massaging the predictions so that they match up with the other models.

Also, the Stan code currently won't accept sites with missing environmental data, so we'll either have to figure out why some site/year combinations are missing from the environmental data (#64) and fix them, or impute values for the missing cases.

davharris:
Here's an example of how the predictions look.  Colored points connected by red lines are observations by different observers (colors).  Black/gray lines are posterior samples over estimates for what the median observer would find.

Purple lines are 95% prediction intervals, taking into account what we knew about each observer in 2003.  So when Ms. Cyan stops visiting the site, the intervals expand to reflect our ignorance about the biases of Mr. Green and Dr. Blue.

image

ethanwhite:
@davharris - this is looking great. Nice work! Let me know when you're finished with the additional cleanup and integration and then I'll give everything a final read through and we can get this in.

davharris:
The current breakdown of variance in observed richness according to the model is currently something like:

* 35% variation in "anchor" position.

* 45% autorcorrelated fluctuations around the "anchor" position

* 15% variation among observers

* 5% uncorrelated observation noise

The "anchor" is mainly determined by NDVI, and is mostly constant because NDVI is mostly constant within sites.

davharris:
Looks like I mostly fixed the MCMC problems in 8de38f6 and package version problems in the subsequent 4 commits.  Things are mixing pretty well now, for the most part and I think the parameter estimates are making sense.

Part of the problem is the tight relationship between two of the autoregression parameters shown below ("beta", where higher values mean less mean reversion; and "sigma",  where higher values mean more noise between time steps).  Any combination along the ridge in the below graph is plausible, although yellower points have higher posterior probability.

image

As I was poking around, I set a prior that yanked things to the left more than I probably needed to, which could be responsible for some of the remaining MCMC problems (the left side also has lower posterior probability).  Overnight, I'm going to try a weaker prior and also see what happens if I put NDVI back in as a predictor.

davharris:
Still not working great.  I think this might be part of the problem, although I don't understand it yet.  The x-axis is the long term variance within sites and the y-axis is the variance among site means. My current hypothesis is that the sampler is having trouble with this banana shape.  No idea what to do about it yet though...

image

sdtaylor:
fyi this model has gotten pretty complex, and aside from any issues with the input data, it's doubtful I'd be able to judge whether it's performing correctly or help diagnose any issues with it.

davharris:
It's actually gotten a bit simpler recently, but I completely agree that nobody except me should be expected to track all the massive changes I'm making. Ethan offered to go over it with me once I've settled down and are ready to consider merging. So I think you're off the hook.

Thanks though!

davharris:
replacing with 

`lmer(richness ~ (1|site_id) + (1|observer_id) + (1|year)`

😢

PR #63, opened 2016-08-10
Fix min num years

sdtaylor:
`get_richness_ts_env_data()` wasn't enforcing `min_num_years`. see #62

### DIFF
--- a/R/forecast-bbs-core.R
+++ b/R/forecast-bbs-core.R
@@ -117,9 +117,7 @@ get_bbs_data <- function(start_yr, end_yr, min_num_yrs){
     bbs_results <- dbSendQuery(con, bbs_query)
     bbs_data <- dbFetch(bbs_results) %>%
       filter_species() %>%
-      filter(year >= start_yr, year <= end_yr) %>%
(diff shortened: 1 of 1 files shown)


R/forecast-bbs-core.R
```diff
@@ -117,9 +117,7 @@ get_bbs_data <- function(start_yr, end_yr, min_num_yrs){
     bbs_results <- dbSendQuery(con, bbs_query)
     bbs_data <- dbFetch(bbs_results) %>%
       filter_species() %>%
-      filter(year >= start_yr, year <= end_yr) %>%
       group_by(site_id) %>%
-      filter(min(year) == start_yr, max(year) == end_yr, length(unique(year)) >= min_num_yrs) %>%
```
ethanwhite: Now that these have been removed it looks like maybe we don't need them as arguments to the function? In that case we should take them out of the function definition and any associated function calls.

sdtaylor: k, got rid of those.

ethanwhite:
Thanks @sdtaylor. Looks great.

Issue #61, opened 2016-08-09
Incorporate observer effects


davharris:
I'd feel better if I knew what was going on with big richness jumps like the one plotted below (site_id 27057).

![image](https://cloud.githubusercontent.com/assets/843017/17531609/ab2c577c-5e4a-11e6-94da-bde45bb2e014.png)

Is that a _real_ change in richness, or did a new observer just start counting the birds in 1993? These big jumps have a big impact on the tails of the random walk, so I think we'll want to see if we can understand them better.

My RPostgres isn't behaving at the moment, so I can't easily access the observer data right now.

sdtaylor:
@ethanwhite has mentioned funny things around this time period too when using the env data. I thought it might be something to do with species name changes and our filtering of them, but I just checked the raw data downloaded by the retriever and it shows the same thing for this site. A new observer is probably the best explanation. But I will try to do a little detective work on this.

davharris:
Thanks Shawn!

ethanwhite:
Yeah, we should definitely do some digging on some of these rapid transitions. In general I think some time spent just looking through the compiled time-series and getting a feel for what sorts of things (weird and not so weird) are going on will be really useful.

Something else other than observer changes to look out for is any potential changes in the route. In concept I believe any change in the route is supposed to be accompanied by a change in the route number to avoid these issues, but I haven't looked into that in a while and could see it causing similar sorts of changes.

sdtaylor:
For this site it is the observer changing over and counting more birds. here is the data for it from the BBS_weather file.

```
   year    obsn totalspp
1  1985  990081       42
2  1986  990081       38
3  1987  990081       49
4  1988  990081       43
5  1989  990081       39
6  1990  990081       37
7  1991  990081       33
8  1992 1010060       66
9  1993 1010060       63
10 1994 1010060       61
11 1995 1010060       63
12 1996 1010060       64
13 1997 1010060       64
14 1998 1010060       63
```

I'll look around and see how prevalent this is.

davharris:
Thanks Shawn! I can start looking into adding an observer effect to my Stan code if this seems like it becomes a pattern.

Here are the 20 sites with the largest one-year jumps, in case a list is helpful.

61123
17003
91041
27057
91032
66033
72035
52002
46017
7020
72012
61045
58001
92019
81043
61064
58023
14151
91025
91012

sdtaylor:
[...]

PR #68, opened 2016-08-23
add observer to richness timeseries data

sdtaylor:
different observers have a large affect on the observations, see #61. Note that missing observer data will match missing richness data for when a route was not sampled in a particular year.

with the new db_engine() function, adding variables is a breeze.


davharris:
☺

PR #72, opened 2016-09-22
add random effects as predictors

davharris:
This isn't 100% macho-stats approved, since it doesn't propagate the uncertainty about any of these values.  But it should be fine for now.

(also: ugh.  I just noticed that RStudio removed a bunch of whitespace. If that makes this hard to review, I can undo it.)


R/forecast-bbs-core.R
```diff
@@ -299,17 +299,51 @@ add_observers=function(bbs_data){
...
 }
 
+#' Add point estimates for random effects associated with site_id and observer_id
+#' @param df a data.frame, as produced by \link{get_bbs_data}
+#' @param last_training_year the final year to include in the training set for
+#'        fitting the mixed model
+#' @importFrom lme4 lmer ranef
+add_ranefs = function(df, last_training_year) {
+
+  df = add_observers(df)
```
ethanwhite: Any reason not to add this to `get_richness_ts_env_data` and then use the resulting dataframe as the starting point instead of recalculating richness again in a separate location?

ethanwhite:
This looks fine. I've asked one question related to potentially avoiding some code duplication.

R/forecast-bbs-core.R
davharris: nope. Good call

davharris: Oh, I did have a (potentially) good reason.  I'm not using this just for analyses involving richness.  The first place I was planning on using it was for single-species models.  I'll weigh the options and see what makes the most sense.

davharris:
I'll resubmit using #73 to reduce duplication.

PR #73, opened 2016-09-23
deduplicate munging code

davharris:
Since we're going to be comparing species-level and richness-level analyses, it makes sense to use the same code for both so they don't get out of sync.  This also makes #72 simpler.

Quick benchmarking suggests that this code adds 2 seconds to the richness importing (because some munging is done on redundant species-level rows).  If that's a problem, I can optimize better.


davharris:
Actually, I take that back about timing.  Reducing the number of merges seems to have balanced those 2 seconds.

davharris:
By the way, I fixed the "trailing whitespace" issue, so that shouldn't be cluttering up my commits anymore.

R/forecast-bbs-core.R
```diff
@@ -252,32 +252,23 @@ get_pop_ts_env_data <- function(start_yr, end_yr, min_num_yrs){
...
-  bbs_data <- get_bbs_data()
-  
-  site_lat_long = bbs_data %>%
-    select(site_id, lat, long) %>%
-    distinct()
-  
-  richness_data <- bbs_data %>%
+  get_pop_ts_env_data(start_yr, end_yr, min_num_yrs) %>%
+    collapse_to_richness()
+}
```
sdtaylor: :+1:

sdtaylor:
`x=get_richness_ts_env_data(1982,2013,25)`  
When I run this on `master` I get 32670 rows, when I run it on this branch I get 28867. Looks like this is dropping NA richness values somewhere.

davharris:
Oh, I didn't realize we wanted NAs. I'll add them back in. 

Dave

sdtaylor:
I can't find it anywhere, but I do remember you saying that NA richness values can be accommodated in the models.

davharris:
Okay, think it's fixed.  Re-running the ipython notebook to make sure it doesn't break anything.

sdtaylor:
The NA's get filled in, but now instead of NA they are all equal to 1 :confused:

davharris:
Right.  Fix coming soon. 😧

davharris:
Nice catch.  I was counting "NA" rows as one species.  This should do it. Thanks for the careful review!

davharris:
Gah. Just take my computer away.  Fix coming soon...

davharris:
okay, ready for review... sorry...

sdtaylor:
Cool, looks good to me.

Before merging though. The random effects stuff from #74 snuck in here. Should I just merge it with this or put it in the other PR?

davharris:
whatever you think is best.  Sorry for apparently leaving my brain at home this morning...

PR #74, opened 2016-09-23
add random effects (depends on #73)

davharris:
This is a revised version of #72, in light of @ethanwhite's comments on code duplication.


davharris:
merged in #73

PR #75, opened 2016-09-23
analyze mixed models & residuals

davharris:


davharris:
Looks like the stan model is pretty uncertain about most of the observer effects: `observer_id` only explains 32% of the variance in the `observer_effect` values across MCMC samples. In other words, finding out an observer's ID code typically reduces our uncertainty in their bias by 1/3.

I think that puts us squarely in the range where a) the observer effects are strong enough that we can't ignore them and b) we also can't ignore the uncertainty.  So it's good that we've been planning to propagate the errors through.

I've attached an image of what 30% looks like:

image

The dark curve is the approximate density for *all* the observers across all MCMC samples.  The colorful curves each show the posterior density for a single observer.  They do tend to be narrower than the full distribution, but not by a whole lot.

Code to create the figure and calculate R-squared:

```
colors = alpha(dichromat::colorschemes$Categorical.12, .75)

x = obs_model$observer_effect

plot(density(x), yaxs = "i", ylim = c(0, .25), lwd = 0, xlab = "observer deviate (# species)")
for(i in sample.int(ncol(x), 100)){
  lines(density(x[,i]), col = colors[1 + (i %% length(colors))])
}
lines(density(x, n = 1000), lwd = 5)

#Quick/dirty R-squared
1 - var(colMeans(x)) / var(c(x))
```

davharris:
@ethanwhite Let's call this ready.  The history is messy and I wouldn't object to any squashes you think would be wise.

ethanwhite:
Thanks Dave. Looks good. I'm just going to go ahead and squash and merge and whole thing since it's basically just one cohesive change.

PR #76, opened 2016-09-27
single-species SDMs

davharris:
(replace #45)

I'll upload the shear maps as a separate PR.


sdm.R
```diff
@@ -0,0 +1,95 @@
...
+  # Determine the number of trees to include by using the first portion of
+  # the training set to predict the remainder of the training set.
+  g = gbm(
+    my_formula,
+    data = xy,
+    distribution = "bernoulli",
+    interaction.depth = interaction.depth,
+    train.fraction = mean(xy$year < first_validation_year),
+    n.trees = max_n_trees
+  )
```
sdtaylor: it's stopping here for me with this error. It probably needs #71?

 `Error in eval(expr, envir, enclos) : object 'start_time' not found`

sdm.R
```diff
@@ -0,0 +1,95 @@
...
+    distribution = "bernoulli",
+    interaction.depth = interaction.depth,
+    train.fraction = 1,
+    n.trees = n.trees
+  )
+  
+  p = predict(g_full, test_x, n.trees = n.trees, type = "response")
+  
+  write.table(p, file = paste0("sdm_predictions/", sp_id, ".txt"),
+              row.names = FALSE, col.names = FALSE)
```
sdtaylor: Instead of writing results files for each species, can all the results be summarized into site/year richness estimates?

davharris: Yes, I agree that's better for the final product.  Right now, I just wanted to write the results as I went so that I won't be empty-handed if/when my job fails 12 hours in. I'll change it back once I've run it all the way through Hipergator.

sdm.R
davharris: Oh, right. I'll remove that for now. Thanks

sdm.R
```diff
@@ -0,0 +1,95 @@
...
+
+start_yr = 1982
+end_yr = 2013
+min_num_yrs = 25
+last_train_year = 2003
+
+# GBM parameters
+first_validation_year = 2000 # for temporal cross-validation
+interaction.depth = 8
+max_n_trees = 2E4
```
sdtaylor: 20,000 trees is probably a bit much. I'd be willing to bet the accuracy improvements after ~2000 trees are probably < 0.01.  Cutting this down would cut down the processing time by a bunch.

sdtaylor:
Were you looking to merge this now or after it's run thru hipergator and giving richness estimates?

sdm.R
davharris: Yeah, the number of trees depends on the size of the data set and the learning rate/shrinkage value. We can play with this and maybe use gbm.more for the few species that need this many.

davharris:
It finished up early this morning, so I can confirm that it gives richness estimates. Let's see about merging after I address your comments above. Thanks for the detailed review!

davharris:
@sdtaylor I think this addresses all your suggestions.  Want to review the changes so I can send this off to Hipergator? Thanks again!

sdm.R
```diff
@@ -0,0 +1,124 @@
...
+  while (!done_adding) {
+    n.trees = gbm.perf(probe_model, method = "test", plot.it = FALSE)
+    
+    # We're done adding trees when the optimal number of trees stops going up.
+    if (n.trees < (probe_model$n.trees - n_trees_to_add)) {
+      done_adding = TRUE
+    } else{
+      probe_model = gbm.more(probe_model, n.new.trees = n_trees_to_add)
+    }
+  }
```
sdtaylor: I think you are trying to iteratively increase the num_trees to some optimum, but I'm pretty sure this could end up as an infinite loop in some cases.

sdm.R
```diff
@@ -0,0 +1,124 @@
...
+    train_predictions = predict(probe_model, xy, n.trees = n.trees, 
+                                type = "response"),
+    test_predictions = predict(prediction_model, test_x, n.trees = n.trees, 
+                               type = "response"),
+    importances = summary(prediction_model, plot = FALSE),
+    n.trees = n.trees
+  )
+}
+
+results = mclapply(species_ids, fit_species,
```
sdtaylor: fyi foreach from the doParallel packaged can return a single dataframe from a parallel job.

sdtaylor:
On choosing the number of trees. I picked 3 random species and got their score traceplots from gbm.perf. I think we can get away with setting n.trees initially to 5000 ( or maybe a bit higher) and running gbm.perf on that. 

oob1
oob2
oob3

sdm.R
davharris: That seems almost impossible to me, but it certainly can't hurt to add an upper bound.

sdm.R
davharris: good to know

davharris:
It's hard to tell how much we can infer from those plots, given that it's only three species and that the package warns that OOB might underestimate the number of trees.  I think the current strategy + an upper bound on the number of trees will be okay.

Thanks again

davharris:
Okay, I've capped the number of trees to avoid the possibility of (nearly) infinite loops and I'm now saving the validation error curve for each species, so we can see if we need to optimize anything further.

sdtaylor:
Cool, this should run ok. Definitely set the time cap on the hipergator pretty high.

ethanwhite:
@sdtaylor - Thanks for the awesome reviewing. Feel free to go ahead and merge when ready.

PR #77, opened 2016-09-27
shear map in new 'analysis' folder

davharris:
(also remove some dead folders I created last year). This replaces the shear maps in #45


ethanwhite:
Looks good. Thanks!

My one question is whether to keep this in it's own file (which I think should then probably be in `./R/` or put it inside `forecast-bbs-core.R`. My natural inclination is the latter, but I if that inclination to put lots of things in a single file is my Python background talking I'm certainly happy to do the more R-like thing.

davharris:
I think putting it somewhere in the R folder does make more sense.  We'll
probably end up adding ggplot as a dependency eventually anyway.  Good call

davharris:
This should be ready to merge, I think.

PR #78, opened 2016-09-27
manuscript outline

davharris:


ethanwhite:
Looks like a good starting point to me. In it goes.

PR #84, opened 2016-10-13
some comments and added citations

sdtaylor:
I put in a few comments, and also added a separate file to collect citations and how they are referenced. I used this idea in my other manuscript and think it works really well.


ethanwhite:
:+1:

PR #86, opened 2016-10-31
manuscript: update methods/models

davharris:


davharris:
Merging because @ethanwhite already saw/approved the changes.

PR #87, opened 2016-11-07
methods, mainly

davharris:


davharris:
I'll accept this PR because Ethan has already reviewed the changes

PR #88, opened 2016-12-08
WIP: Fit models with observer effects

davharris:

PR #92, opened 2017-01-16
fit observer stan model (solves #61)

davharris:
partially replaces #88


ethanwhite:
Sorry for taking so long on this. LGTM!

PR #93, opened 2017-01-16
Fit Richness Models

davharris:
Supersedes #88.  Depends on #92.  Should probably be squashed before merging.


ethanwhite:
Can you resubmit this exact same branch in a new PR. GitHub doesn't update in response to the merge of the other branch so this will give me a clean look at this PR. Thanks.

ethanwhite:
Reopening since this PR is actually in good shape. I'll close #95 next.

davharris:
I think I fixed this one by merging upstream into the branch. And then didn't realize it was fixed because there were still so many commits. Sorry for the confusion

ethanwhite:
Given the size of the additions my review was necessarily a fairly light weight check, but at that level I didn't see any issues. In it goes!

PR #94, opened 2017-01-25
JSDM

davharris:
Depends on my other two PRs


davharris:
This works, but I *still* can't get it to run in parallel on Hipergator. Running serially now. Ready for code review/discussion/tips on what I might be doing wrong.

sdtaylor:
do you have any errors from the hipergator?

davharris:
It just sits there indefinitely without producing output when I run the code as-is, but it does save my results to the `mistnet_output` folder when I run it with `lapply` instead of `mclapply`

sdtaylor:
does mistnet parallize at all? I'm running `fit_mistnet()` right now without mcapply and it's cpu usage is 200%, which usually means it's maxing out 2 cores.

davharris:
That's really weird.  None of the code I wrote for mistnet is in parallel, although I suppose it's always possible that I'm calling some other parallel code? Maybe the armadillo library?

Thanks for looking into this...

davharris:
On my Mac, I'm only using 100%. When you run it, does it create two process each running around 100%, or does it just have one process going at 200%?

sdtaylor:
On serenity, which is ubuntu, it has one process going at 100%, but on hipergator when I use a dev session it will go to 200%

davharris:
😡😡😡

Dave

sdtaylor:
using a `doParrallel::foreach()` implementation  on hipergator seems to be working, I'll let you know if it actually finishes on a small subset. 

```
library(doParallel)
cl=makeCluster(4)
registerDoParallel(cl)

output = foreach(i=1:8) %dopar% {
    source("mistnet.R")
    fit_mistnet(iter = i, use_obs_model = TRUE)
    }
```

davharris:
Thanks Shawn!

sdtaylor:
Confirmed, using `foreach()` as above works.

mistnet.R
```diff
@@ -0,0 +1,184 @@
...
+  
+  settings = yaml::yaml.load_file("settings.yaml")
+  
+  # Collect data for this iteration, then discard the rest to save
+  # memory
+  obs_model = readRDS("observer_model.rds")
+  data = filter(obs_model$data, iteration == iter) %>% 
+    left_join(bbs, c("site_id", "year"))
+  rm(obs_model)
+  gc()
```
ethanwhite: You don't need to do this here since this is working, but there are also a couple of other ways that never read the full file into memory, e.g., https://stackoverflow.com/questions/23197243/how-do-i-read-only-lines-that-fulfil-a-condition-from-a-csv-into-r. The `sqldf` approach seems particularly light weight and is apparently pretty faster than the alternatives.

mistnet.R
```diff
@@ -0,0 +1,184 @@
...
+  for (i in 1:ncol(x)) {
+    mean_i = mean(x[data$in_train, i])
+    sd_i = sd(x[data$in_train, i])
+    x[,i] = (x[,i] - mean_i) / sd_i
+  }
+  
+  y = data %>% 
+    select(matches("^sp[0-9]+$")) %>% 
+    select(which(colSums(.) > 0)) %>% 
+    as.matrix()
```
ethanwhite: Maybe add a quick comment on what these filters are doing (I can't follow them just from reading).

davharris:
Looks like @sdtaylor's idea didn't solve the problem: instead of fitting 1 model per hour, I'm now fitting 16 models every 16 hours.

I'm going to cancel everything and just type `sbatch mistnet.job` into the terminal sixteen times 🛠

sdtaylor:
how long does each model normally take to run?

davharris:
One hour on one thread. So I should be getting sixteen per hour, not sixteen every sixteen hours. 

Dave

sdtaylor:
So is `dopar` working as expected with mistnet now? What did you change?

davharris:
This should finally be ready for review.  Sorry for the giant mess of 50+ commits...

ethanwhite:
Great, no worries. With the GitHub Squash and Merge option that's easy to fix.

I don't have time for a detailed read through before I leave, but I did give it a quick once over and didn't see any obvious issues. @sdtaylor why don't you take a look and when you're happy go ahead and merge.

ethanwhite:
I don't know enough about the mistnet API to say this looks right, but I didn't see anything that looked problematic, so in it goes.

PR #95, opened 2017-02-06
Models2

davharris:
Supersedes #93, which supersedes #88.

Should probably be squashed due to several false starts.  Sorry for the messy history.


davharris:
@ethanwhite Doesn't look like the history is any cleaner this way? Sorry if I did something silly.

ethanwhite:
Actually, maybe they've updated something and I just didn't notice given that there's a lot going on here. Are any of the commits from the PR I merged this morning listed here or is this just the new commits for this PR?

ethanwhite:
I feel like the diff looks different actually. I wonder if it just took a few minutes to get updated after the merge (which would be a great change in the GitHub interface if that happened).

davharris:
Oh, I think you're right.

ethanwhite:
Yeah, it's definitely clean now and I'm pretty sure it wasn't when I first look. Sorry for the confusion. I'm on it.

ethanwhite:
Closing since #93 was actually fine. GitHub just keeps getting better.

PR #96, opened 2017-02-14
Draft of Introduction

ethanwhite:
This includes the Intro itself, a box that holds the list of things we've been talking about as good practices, and and initial try at a title.

### DIFF
--- a/manuscript/draft.Rmd
+++ b/manuscript/draft.Rmd
@@ -166,3 +166,83 @@ None of the models are true forcasts as they use observed predictor variables (e
 * Need predictions at multiple time scales
 * Prediction quality at short time scales might not be a good indicator of 
   long-term effectiveness
+
+
+
+## Box 1: Ten simple rules for making and evaluating ecological forecasts
+
+### 1. Compare multiple modeling approaches
+
+Typically ecological forecasts use one modeling approach or a small number of
+related approaches. By fitting and evaluating multiple modeling approaches we
+can learn more rapidly about the best approaches for making predictions for a
+given ecological quantity. This includes comparing process based and data-driven
+models.
+
+### 2. Use time-series data when possible
+
+Forecasts describe how systems are expected to change through time. While some
+areas of ecological forecasting focus primarily on time-series data, others
+primarily focus on using spatial models and space-for time substitutions. Using
+ecological and environmental time-series data allows the consideration of actual
+dynamics from both a process [@treddenick2016a] and error structure perspective.
+
+### 3. Pay attention to uncertainty
+
+Understanding confidence in a forecast is just as important as understanding the
+average or expected outcome. Failing to account for uncertainty can result in
+overconfidence in highly uncertain outcomes leading to poor decision making and
+erosion of confidence in ecological forecasts. Models should explicitly include
+sources of uncertainty where possible and evaluations of forecasts should assess
+the accuracy of uncertainties as well as point estimates [@dietz2017].
+
+### 4. Use predictors related to the question
+
+Many ecological forecasts use data that is readily available and easy to work
+with. While ease of use is a reasonable consideration it is also important to
+include predictor variables that are expected to relate to the ecological
+quantity being forecast and dynamic time-series of predictors instead of
+long-term averages. Investing time in identifying and acquiring better predictor
+variables may have at least as many benefits as using more sophisticated
+modeling techniques.
+
+### 5. Assess how forecast accuracy changes with time-lag
+
+In general the accuracy of forecasts decreases with the length of time into the
+future being forecast [@petch2015]. This decay in accuracy and associated
+forecast horizons should be considered when evaluating forecasts and comparing
+models.
+
+### 6. Include an observation model
+
+Ecological observations are influenced by both the underlying processes and how
+the system is sampled. When possible forecasts should model the factors
+influencing the observation of the data.
+
+### 7. Validate using hindcasting
+
+To evaluate the expected accuracy and uncertainty of forecasts assess the
+performance of these forecasts within existing time-series data.
+
+### 8. Publicly archive forecasts
+
+To allow the future evaluation of the accuracy and uncertainty of forecasts the
+forecast values and/or models should be archived so that they can be assessed
+after new data is generated [@mcgill2012]. Enough information should be provided
+to allow an unambiguous assessment of the forecast performance.
+
+### 9. Make short-term and long-term predictions
+
+In cases where long-term predictions are the primary goal, short-term should
+also be made to accommodate the time-scales of planning and management decisions
+and to allow the accuracy of the forecasts to be quickly evaluated
+[@treddenick2016a];
+
+### 10. Compare forecasts to simple baselines
+
+Compare the accuracy of the forecasts to simple baselines such as the
+time-series average to determine if the modeled forecasts are more accurate than
+the naive assumption that the world is static.
+
+
+
--- a/manuscript/draft.Rmd
+++ b/manuscript/draft.Rmd
@@ -4,44 +4,55 @@ Authors: Harris, Taylor and White?
 
 ## Introduction
 
-Quality of predictions is important:
-  * helps determine policy/set priorities
-  * something something climate change
-  * need to know both accuracy of point estimates and how reasonable our prediction intervals are.
-  * Overconfident predictions can be worse than uncertainty
-  * Quality of prediction is one metric for how much we know about ecological processes
-Unclear how much to trust existing predictions:
-  * Strong assumptions (see "misc" below)
-  * many *interesting* predictions
-  * Potential problems with confidence intervals
-  * Potential disconnect between training target and actual goals
-  * Difficult to test model assumptions/predictions because of time scales
-    * Most scientists making predictions today will have retired by 2050
-Demonstrate best practices \& their contribution to better predictions
-  * Use long-term data sets such as BBS to evaluate predictive skill
-  * Note the large change in average temperatures since the 80's; comparable to 
-    the potential changed in the next 20 years 
-  * Lots of talk about how we should do it, fewer examples of doing them all together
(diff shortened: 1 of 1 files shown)


manuscript/draft.Rmd
```diff
@@ -166,3 +177,83 @@ None of the models are true forcasts as they use observed predictor variables (e
...
+long-term averages. Investing time in identifying and acquiring better predictor
+variables may have at least as many benefits as using more sophisticated
+modeling techniques.
+
+### 5. Assess how forecast accuracy changes with time-lag
+
+In general the accuracy of forecasts decreases with the length of time into the
+future being forecast [@petch2015]. This decay in accuracy and associated
+forecast horizons should be considered when evaluating forecasts and comparing
+models.
```
sdtaylor: Also important because the most accurate model may not be the same at every time step. Ward et al 2014 has some subtle examples of this in their figures.

ethanwhite: Thanks. I'll add some language. Do Ward et al. discuss this or is it something you picked up on without them mentioning it?

sdtaylor: They don't mention it. In their case the best model never changes, but the graphs do show relative rankings sometimes changing with lead time.

ethanwhite: Thanks. I just added some language. Let me know if that does the trick.

davharris:
This is excellent.  I'm just going to go through the deleted outline and make sure no babies were thrown out with the bathwater, then I'll merge.  Thanks!

manuscript/draft.Rmd
```diff
@@ -166,3 +177,84 @@ None of the models are true forcasts as they use observed predictor variables (e
...
+
+Forecasts describe how systems are expected to change through time. While some
+areas of ecological forecasting focus primarily on time-series data, others
+primarily focus on using spatial models and space-for time substitutions. Using
+ecological and environmental time-series data allows the consideration of actual
+dynamics from both a process [@treddenick2016a] and error structure perspective.
+
+### 3. Pay attention to uncertainty
+
+Understanding confidence in a forecast is just as important as understanding the
```
davharris: Would be good to make the point (here, or possibly in the main text instead) that overconfidence is/can be much worse than under-confidence

manuscript/draft.Rmd
```diff
@@ -1,47 +1,58 @@
...
+processes, model choice, and forecast evaluation [@dietz2017]; and 3) conducting
+meaningful evaluations of the forecasts by hindcasting, archiving short-term
+forecasts, and comparing forecasts to baselines to determine if the forecasts
+are more accurate than assuming the system is basically static [@Perretti2013].
+
+In this paper we attempt to forecast the species richness of breeding birds at
+hundreds of locations throughout North America while following the best
+practices in Box 1 for ecological forecasting. To do this we combine 30 years of
+time-series data on bird distributions with monthly time-series of climate data
+and satellite-based remote-sensing. We compare traditional distribution modeling
```
davharris: This might be a good place to note the large-ish increase in average temperatures over this time period, comparable to what's expected in the next ~20 years

manuscript/draft.Rmd
```diff
@@ -1,47 +1,58 @@
...
+then used to predict where each species will occur in the future and the
+probabilities of occurrence are summed across species to predict future species
+richness [e.g., @thomas2004]. Alternatively, models that directly relate spatial
+patterns of species richness to the environment have been developed and
+generally perform equivalently to species distribution modeling based methods
+[@algar2009, @distler2015].
+
+However, despite the emerging interest in forecasting species richness and other
+aspects of biodiversity, little is known about how effectively we can anticipate
+their dynamics. What we do know comes from a small number of hindcasting
```
davharris: Do we want to be more explicit that evaluation of long-term forecasting is hard because of time lag before we find out how well we did?

manuscript/draft.Rmd
```diff
@@ -1,47 +1,58 @@
-# Title
+# Moving towards large-scale data-intensive forecasting of species richness in breeding birds
```
davharris: How about something more like

*Best practices for forecasting biodiversity changes: an example with breeding birds*

I feel like my version does a better job of emphasizing that the paper is about best practices, not about breeding birds specifically. Also, I think "best practices" is less wishy-washy than "moving towards".  More like, "we know what to do, let's start doing it" and less like "I guess we should form a committee to figure out the way forward."

manuscript/draft.Rmd
```diff
@@ -166,3 +177,84 @@ None of the models are true forcasts as they use observed predictor variables (e
...
+
+Typically ecological forecasts use one modeling approach or a small number of
+related approaches. By fitting and evaluating multiple modeling approaches we
+can learn more rapidly about the best approaches for making predictions for a
+given ecological quantity. This includes comparing process based and data-driven
+models.
+
+### 2. Use time-series data when possible
+
+Forecasts describe how systems are expected to change through time. While some
```
davharris: Do we want to distinguish here between time series of richness vs time series of environmental predictors? Emphasize that both are good to have?

manuscript/draft.Rmd
```diff
@@ -166,3 +177,84 @@ None of the models are true forcasts as they use observed predictor variables (e
...
+to allow an unambiguous assessment of the forecast performance.
+
+### 9. Make short-term and long-term predictions
+
+In cases where long-term predictions are the primary goal, short-term should
+also be made to accommodate the time-scales of planning and management decisions
+and to allow the accuracy of the forecasts to be quickly evaluated
+[@treddenick2016a];
+
+### 10. Compare forecasts to simple baselines
```
davharris: Maybe this could be folded into 1? Like, multiple models, *including simple baselines*? This would make room for the extra item I added at the top of the box/list

manuscript/draft.Rmd
```diff
@@ -166,3 +177,84 @@ None of the models are true forcasts as they use observed predictor variables (e
 * Need predictions at multiple time scales
 * Prediction quality at short time scales might not be a good indicator of 
   long-term effectiveness
+
+
+
+## Box 1: Ten simple rules for making and evaluating ecological forecasts
```
davharris: We're missing spatial/site level stuff, I think.  I think something like 30% of the variance in species richness was site-level stuff that couldn't be explained by annual fluctuations, observer effects, or the environmental variables we included.  This would be a good place to talk about mixed models, latent variables, JSDMs, spatial random effects, etc.

manuscript/draft.Rmd
davharris: Also, this might be a good place to mention error propagation (i.e. our uncertainty in one step should affect any estimates made subsequent steps). Otherwise, we can put it somewhere in the main text.

manuscript/draft.Rmd
```diff
@@ -1,47 +1,58 @@
...
-    * How much to trust the predictions
+Forecasting the future state of ecological systems is increasingly important for
+planning, management, and testing the performance of ecological models
+[@clark2001, @dietz2017]. One area of particular interest for forecasting is
+biodiversity since it is important for ecosystem function, central to
+conservation planning, and expected to be influenced by anthropogenic global
+change. Anticipating potential changes in biodiversity is crucial for large
+scale management and conservation, and for addressing debates regarding whether
+local scale diversity is declining and may continue to decline in response to
+anthropogenic influences [@dornelas2014].
```
davharris: This might be a good place to mention that quality of prediction is one metric for how much we know about ecological processes. If we can't predict, we [or at least our models] clearly don't understand what's going on. So this isn't just important for applied folks.

manuscript/draft.Rmd
ethanwhite: I'm not happy with the title either. I'm not convinced that the current introduction quick backs up this alternative. I'd propose that we see how the paper shapes up and then tinker with the title.

ethanwhite: Thoughts on:

*Using best practices for forecasting biodiversity changes in breeding birds*

as a middle ground?

manuscript/draft.Rmd
ethanwhite: I've expanded the "testing the performance of ecological models" text to try to better capture this.

manuscript/draft.Rmd
ethanwhite: Adding this sounds good, but I'm not sure exactly how to phrase this. Can you propose some language for it?

manuscript/draft.Rmd
ethanwhite: The current language does split them out "Using ecological and environmental time-series data allows..." I'm open to suggestions for additional changes, just not sure exactly what you're looking for. We want to keep things concise so that the box will fit on a single page, but I could see changes to "...allows the consideration of actual dynamics from both a process [@treddenick2016a] and error structure perspective." that might make the distinct importance of the two kinds of data clearer.

manuscript/draft.Rmd
ethanwhite: > Would be good to make the point (here, or possibly in the main text instead) that overconfidence is/can be much worse than under-confidence

This what I was trying to accomplish with "Failing to account for uncertainty can result in
overconfidence in highly uncertain outcomes leading to poor decision making and
erosion of confidence in ecological forecasts." i.e., there are serious consequences for overconfidence. I'm certainly open to suggested changes to how this is communicated. I agree we should also discuss this in the main text, specifically in the Discussion both general and in the context of the fact that we aren't forecasting environmental conditions yet.

> Also, this might be a good place to mention error propagation (i.e. our uncertainty in one step should affect any estimates made subsequent steps). 

I've mentioned it briefly here

ethanwhite:
Thanks for the great comments @davharris. I just pushed a new commit with a bunch of changes and responded to a few things in the comments. Back to you.

manuscript/draft.Rmd
davharris: That seems good for now.

PR #98, opened 2017-03-07
include gimms version number 0

sdtaylor:
The gimms package was recently updated to include version 1 of the gimms dataset and set version 1 as the default. Version 1 changes the file structure so most of get_ndvi_data will have to be redone if we switch to it. This will stick to version 0 for now.


ethanwhite:
Thanks @sdtaylor. Does version one have later dates than version 0?

sdtaylor:
Yes it does. 0 goes to 2013 and 1 goes to 2015, so it's something we'll want to update to.

PR #100, opened 2017-03-10
Fix gimms stuff

sdtaylor:
more minor fixes to the updated gimms package. this should be it unless they have another major update.


ethanwhite:
Thanks @sdtaylor

PR #108, opened 2017-05-11
first go at data

sdtaylor:


manuscript/draft.Rmd
```diff
@@ -81,53 +83,57 @@ in the future.
...
     `r settings$last_train_year`
-  * Sites were included in the analysis if there were at least 
+  * Sites were included in the analysis if there were at least
     `r settings$min_num_yrs` years of data available during this period.
-  * The models were tested/evaluated on the data from 
-    `r settings$last_train_year + 1` `r settings$end_yr`.
+  * The models were tested/evaluated on the data from
+    `r settings$last_train_year + 1` `r settings$end_yr`.  
+
+  Bird species richness was compiled from the Breeding Bird Survey (BBS) obtained using the Data Retriever software [@pardieck2017, @morris203]. The BBS data was initially filtered to exclude all aquatic species, unidentified species, and hybrids. All data from routes that did not meet BBS acceptable criteria were excluded. Only routes within the continental U.S.A were used so as to match the climatic dataset.
```
ethanwhite: "aquatic species..." -> "nocturnal, crepuscular, and aquatic species (since these species are not well sampled by BBS methods), as well as unidentified...".

manuscript/draft.Rmd
```diff
@@ -81,53 +83,57 @@ in the future.
...
+  Nineteen bioclimatic variables were derived from mean monthly temperature and precipitation from the 4km resolution PRISM dataset, allowing us to have yearly values for each variable [@prism]. This was subset to the eight variables identified by Harris [-@harris2015] to be non-correlated. For each BBS route the mean value for each variable within a 40km radius, equal to the length of a BBS route, was used to account for spatial heterogeneity. Satellite derived NDVI was from the NDIV3g dataset and was available from the time period 1981-2015 [@pinzon2014]. Elevation was from the SRTM 90m elevation dataset obtained using the R package raster [@hijmans2014, @jarvis2008].
```
ethanwhite: "Nineteen bioclimatic variables were derived from mean monthly temperature and precipitation from the 4km resolution PRISM dataset, allowing us to have yearly values for each variable [@prism]. This was subset to the eight variables identified by Harris [-@harris2015] to be non-correlated. "
-->
"We used the 4 km resolution PRISM data [@prism] to calculate eight bioclimatic variables identified as relevant to bird distributions [@harris2015] for each year in the dataset: X, Y, Z" where X, Y, Z is a list of the variables we used.

ethanwhite: We should describe the resolution that all of the data sources are used at and that these are based on the starting coordinates for each route.

manuscript/draft.Rmd
```diff
@@ -81,53 +83,57 @@ in the future.
...
+  All models described below were fit using data from the years 1982-2003, and evaluated using data from the years 2004-2015. The year 1982 corresponds to the first complete year of NDVI observations. Only BBS routes with at least 25 yearly observations in the years 1982-2015 were used.
```
ethanwhite: "All models described below were fit using data from the years 1982-2003, and evaluated using data from the years 2004-2015. The year 1982 corresponds to the first complete year of NDVI observations. "
-->
"Data from 1982 (the first complete year of NDVI data) to 2003 were used for training models and data from 2004 to 2015 was used for testing."

ethanwhite:
LGTM for a first draft. @davharris take a look when you get a chance and merge when you're happy.

settings.yaml
```diff
@@ -4,14 +4,18 @@ min_num_yrs: 25
...
 vars: 
-  - bio2 
-  - bio3 
-  - bio5 
-  - bio8 
-  - bio9
-  - bio15 
-  - bio16 
-  - bio18 
+  - bio2 #Mean Diurnal Range (Mean of monthly (max temp - min temp))
```
davharris: I like these comments.  Should the comments at the very bottom for 5,6,7 be there, or are they vestigial?

manuscript/draft.Rmd
```diff
@@ -81,53 +83,57 @@ in the future.
...
+  We used the 4 km resolution PRISM data [@prism] to calculate eight bioclimatic variables identified as relevant to bird distributions [@harris2015] for each year in the dataset: mean diurnal range, isothermality, max temperature of the warmest month, mean temperature of the wettest quarter, mean temperature of the driest quarter, precipitation seasonality, precipitation of the wettest quarter, precipitation of the warmest quarter. Satellite derived NDVI with an 8 km resolution was from the NDIV3g dataset [@pinzon2014] and was available from the time period 1981-2013. Elevation was from the SRTM 90m elevation dataset [@jarvis2008] obtained using the R package raster [@hijmans2014]. BBS route starting point coordinates were used to extract all environmental data. Climatic variables and NDVI were aggregated to the average values within a 40 km radius of the start point, equal to the length of a BBS route. NDVI was also subset to average summer (May, June, July) and winter (December, January, Feburary) values.
```
davharris: If I remember correctly, we had a rule for shifting the Fall climate data back one year. For instance, Fall 1998 happened after the 1998 transect, so we used Fall 1997. If that's correct, I think we should have a sentence about it.

settings.yaml
sdtaylor: if your curious about bio3, you need to also know 5, 6, and 7

davharris: Ah, good thinking.  Maybe just add a comment saying that above 5-7.

sdtaylor:
@davharris  I saw the comment about seasons in my email but not here, did you delete it?

davharris:
No, it probably got hidden when you added a new commit. 

The gist was, do we need to say anything about how we handled Fall climate data? Did we do something like using Fall 1998 data in the 1999 transects, on the theory that the 1999 climate hadn't happened yet?

Dave

sdtaylor:
No, I don't remember doing anything like that. I don't see anything in the code that that does anything like that either.

davharris:
Oh, okay. I vaguely remembered talking about something like that a year ago and must have misremembered the outcome of the conversation. 

Now that I think about it some more, I think we decided not to do anything like that because the transact runs occurred over such a long period and it wasn't clear what date to use for the cutoff. 

Given that it's kind of weird to ask "how will next month's (June) weather affect what birds we'll find on today's transect run (in May)", it's probably worth adding a sentence pointing out that quirk of our data. 

Or you could just merge now and we can deal with it later.

Dave

ethanwhite:
Actually I think this might be worth chatting about a bit more. Maybe merge this and open an issue for us to discuss?

sdtaylor:
Sure

Issue #111, opened 2017-05-15
Make figures and tables


ethanwhite:
At our last meetings we planned out the following figures and tables for the ms:

- [x] Table 1: Model methods
- [x] Figure 1: Example time-series w/forecasts
- [x] Figure 2: Observed-predicted plot for 3 time lags
- [x] Figure 3: Time lag results figure (Main results)
- [ ] Figure 4: Violins of fit at 3 time lags
- [ ] Figure 5: Observation model figure

The sketches for each of these figures is below. If you are already working on or want to work on a figure just leave a note in the the comments (and maybe ping the Slack channel) to avoid duplicate work.

![2017-05-11 13 53 27](https://cloud.githubusercontent.com/assets/744427/26079610/55c36e10-3991-11e7-946e-bee53c1e5c5f.jpg)

![2017-05-11 13 53 21](https://cloud.githubusercontent.com/assets/744427/26079449/d0b07d8a-3990-11e7-9a8b-a51269828278.jpg)

New sketch for the observation model figure from 2017-05-18 meeting:

![img_0862](https://cloud.githubusercontent.com/assets/744427/26216380/daa9b516-3bd1-11e7-9920-86a18796efd6.JPG)

davharris:
Table 1: 

|model             |environment |time |target                   |site_id |
|:-----------------|:-----------|:----|:------------------------|:-------|
|Naive             |X           |✔︎    |richness                 |✔︎       |
|Average           |X           |✔︎    |richness                 |✔︎       |
|Auto-ARIMA   |X            |✔︎    |richness                 |✔︎       |
|Random Forest SDM |✔︎           |X    |single species presence  |X       |
|Mistnet JSDM      |✔︎           |X    |full species composition |X       |
|GBM richness      |✔︎           |X    |richness                 |X       |

sdtaylor:
working on fig 5.

sdtaylor:
Figure 2

![image](https://cloud.githubusercontent.com/assets/12282062/26213354/b5b0c85e-3bc6-11e7-9ac1-868df0b822b1.png)

davharris:
I think these are all finalized now.

PR #117, opened 2017-05-18
add plotting script

davharris:
The first 30 lines or so are to make the tidy data frame with all the results I sent over slack. The next chunk is playing with Figure 3 (#111). The rest is probably not useful; it's looking at observer-effect uncertainty.


ethanwhite:
It would be good to get a version of this with the current state of all plotting code merged so that @sdtaylor and I can start helping out with finalizing figures.

ethanwhite:
(oops hit `Ctrl-Enter` a little too fast)... Is this the current state of things or are there changes from the last few weeks that should go in before I merge this?

davharris:
This will need to be updated to accommodate changes I'm making for predicting to 2050. Will have a working version by Monday.

sdtaylor:
Can you change the filenames of the images to match #111, (fig1, fig2, etc)?. 
I can see which ones correspond to figs 1, 2, and 3, but I'm not sure which observer figure is which.

davharris:
Based on some discussions with @ethanwhite, I've made some changes to the figures. The styling & labels still need work in some places, but the general idea is about right.

I've also added them to the manuscript in approximately the right places.  They don't always line up perfectly with the text, because the text hasn't caught up everywhere to the figures.  Hope this helps @sdtaylor

ethanwhite:
Nice work @davharris! I'm going to go ahead and get this merged.

PR #118, opened 2017-05-18
Preliminary results section

davharris:


manuscript/draft.Rmd
```diff
@@ -273,6 +273,54 @@ Looked at how error changed as the time horizon of forecasting lengthened
...
+stacked SDMs' deviance was dramatically higher than the next-worst model even
+though its RMSE was not unusual. On the other hand, the richness-GBM model and
+especially the mistnet JSDM had reasonably well-calibrated uncertainty
+estimates, even in the presence of uncorrected observer biases (Figure 3C). This
+prevented their deviance from rising as quickly as the site-level models when
+the observer population changed (Figure 3B).
+
+In general, the models produced confidence intervals that were too narrow,
+indicating overconfident predictions. Of these, the random forest-based SDM
+stack was the most overconfident, as discussed above. As one might expect,
```
ethanwhite: Drop "As one might expect"

manuscript/draft.Rmd
```diff
@@ -273,6 +273,54 @@ Looked at how error changed as the time horizon of forecasting lengthened
...
+differences among observers. The remaining 9% represents residual variation,
+where a given observer might report a different number of species on different
+runs of the same transect. We found remarkably little autocorrelation in these
+residuals ([[about 0.04, +/- something]]), suggesting richness is close to
+stationary on these time scales, once observer differences are accounted for.
+
+The substantial uncertainty associated with differences among observers caused
+reliable improvements in both mean-squared-error and deviance when observer
+effects were controlled for (Figure 4). This improvement was smallest for the
+"Naive" model, as discussed below.
```
ethanwhite: I think this paragraph could use a bit more detail in relation to our discussions yesterday. We should talk about how it results in an improvement in the average RMSE and deviance across the board, but that in some cases this is driven more by outliers than the body of the distribution. We should note that even when this is the case it means that the number of really poor forecasts is decreased.

"reliable improvements" indicates to me that that there are consistent improvements across sites. I'd recommend rephrasing during the expansion to avoid confusion.

davharris:
Thanks @ethanwhite!

I think I'll leave this section alone for a bit, since we're going to re-run the analyses soon and will probably get slightly different results anyway.  

I've made sure the "allow edits from maintainers" box is checked: this will help avoid merge conflicts because either of you can easily modify this branch if you want to add something to the results.

ethanwhite:
I'm going to go ahead and merge this since it's a good draft to allow me to start doing a top to bottom walk through of the ms.

manuscript/draft.Rmd
ethanwhite: Implemented up incoming edit.

manuscript/draft.Rmd
ethanwhite: Implemented up incoming edit.

PR #119, opened 2017-05-19
some discussion edits

davharris:
Not 100% sure these are all improvements, but some of it should be useful.


manuscript/draft.Rmd
```diff
@@ -283,83 +283,88 @@ biodiversity in the form of species richness. The results of this effort are
...
-outperforming time-series based approaches at sufficiently long time
-scales. These results suggest that assessment of how forecast performance
-changes with time lag is not only important for determining how far in the
-future to forecast, but also for determining models and processes are most
-relevant for making forecasts at different time scales.
+environmental variables and using the resulting models to make forecasts []. Two
+comparisons of the the SDM and direct richness modeling approaches reported that
+the methods yielded equivalent results for forecasting diversity (some other
+comparisons that found poor SDM performance were [[tainted]] by the rounding
+error alluded to above and are not considered here). While our results generally
```
ethanwhite: As with above my first inclination is to leave out this parenthetical as not central to our message. We could add a 'but see' to the first sentence citations if we think we need to acknowledge the rounding based paper(s), but my temptation here would be to just ignore for them unless someone complains.

The philosophy behind this decision is: they're wrong, so we shouldn't pay attention to the results, nothing we say here will change whether folks use them or not, if we think it's a high priority to address it we should do it directly.

manuscript/draft.Rmd
```diff
@@ -283,83 +283,88 @@ biodiversity in the form of species richness. The results of this effort are
...
+some of the best estimates of uncertainty across all modeling approaches. This
+highlights the importance evaluating models based on their uncertainty as well
+as their mean estimates. Not doing so is particularly problematic in this case
+because the summed species distribution models are overly confident meaning that
+using them would lead to believing that richness would be restricted to a much
+narrower range than would actually occur.
+
+For error metrics that evaluate both models' point estimates and their
+uncertainty, the joint SDM and the nonlinear richness model began to outperform
+the time-series baselines at the longest time-scales assessed. This convergence
```
ethanwhite: Has something changed in the most recent Figure 3 results? The one in the slack discussion shows average and ARIMA still slightly outperforming richness_gbm and mistnet at the longest time scales.

manuscript/draft.Rmd
```diff
@@ -283,83 +283,88 @@ biodiversity in the form of species richness. The results of this effort are
...
+results in simple site-level time-series, which are best predicted by simple
+baselines like the long-term average and the "naive" model.
+
+While stacked species distribution models are the most commonly used approach
+for forecasting the future state of biodiversity [[cite?]], this approach
+provided the worst overall forecasts of all methods evaluated. In fact, the most
+common version of the stacking approach (based on rounding species' occurrence
+probabilities up or down before adding them up) would have led to even worse
+results than what we presented [Pellissier et al. 2013's "probabilistic
+approach" paper, Calabese et al. 2014]. In general, SDM-based forecasts are
```
sdtaylor: SDM's being rarely evaluated isn't entirely true. There are several papers other than rappacciulo2012 that do similar tests. Repeated observations, or testing at yearly timesteps, is a first as far as I can tell.

ethanwhite: @sdtaylor I guess this depends on how we're thinking about rarely. My general impression is that there are on the order of only about 10 dynamic assessments of SDM forecasts, compared to thousands of papers making them. That feels "rare" to me, but if that doesn't seem like the right word to you I'm certainly open to suggestions.

sdtaylor: Yeah, when you put it that way it does seem rare.

manuscript/draft.Rmd
davharris: I'll double-check.

manuscript/draft.Rmd
```diff
@@ -283,83 +283,88 @@ biodiversity in the form of species richness. The results of this effort are
...
-predicted dynamics at each site.
+future, relative to other sites, it generally does. However, the forecasts
+perform relatively poorly at predicting variation in richness over time for any
+given site. If a site is predicted to have a higher richness next year than it
+did last year there is little confidence that this will be the case. This
+results in simple site-level time-series, which are best predicted by simple
+baselines like the long-term average and the "naive" model.
+
+While stacked species distribution models are the most commonly used approach
+for forecasting the future state of biodiversity [[cite?]], this approach
```
davharris: Agreed

manuscript/draft.Rmd
davharris: I'll see if I can put the rounding stuff elsewhere so it doesn't break up the flow.

manuscript/draft.Rmd
davharris: I'll defer to you on this.  I do see a lot of rounding in the literature, which makes it harder to ignore.

ethanwhite: How about a small section in the methods that basically communicates "We don't use rounding. We do this because it's bad. Our results will perform better than any rounding based approaches so the results will be on the high end of those from current approaches to SDM based forecasting"? That way we get the point across in the context of our methodological decision making, but don't have to trip over it in the Discussion?

davharris:
Sounds good. 

Dave

manuscript/draft.Rmd
davharris: Sounds good

davharris:
I've now addressed most of @ethanwhite's comments. The Figure 3 results are about to change soon anyway because we're doing 2 different forecast horizons (and even the horizon we've been using until now will change slightly because of changes in the site inclusion criteria). If there aren't any other problems, I think we can merge?

ethanwhite:
Thanks @davharris!

PR #120, opened 2017-05-24
column name: AOU vs aou

davharris:
Somewhere along the way, the column name for AOU got capitalized by retriever. This makes sure the code still works


davharris:
Added a "WIP" because apparently there are other problems...

davharris:
Okay, got it working. I'm just going to merge because I don't think it requires review.

PR #122, opened 2017-06-06
thoughts on our scale selection

sdtaylor:
I thought we needed something on our choice of yearly timesteps and it turned into this.


ethanwhite:
Looks great. Nice work!

PR #125, opened 2017-06-13
filter years by % training data available (fixes #123)

### LINKED ISSUE #123: filter_ts with shorter training time
Right now, we're filtering the sites so that we have at least 25 observations between 1982 and 2013. Does that need to change if we're only doing 10 years of training? This could give us some sites with only 4 years of training data.

Maybe the criterion should be more like "visited during at least 70% of the training years"? Then we could use the same criterion for both.

As far as I can tell, there's no downside to including a site that doesn't get visited much in the test set, right?

davharris:
I also:

* re-roxygenized everything (to make sure the docs for these functions are up-to-date)
* removed a vestigial SDM script that used these functions


R/forecast-bbs-core.R
```diff
@@ -336,17 +342,22 @@ add_observers = function(bbs_data) {
...
+#' @param min_year_percentage num minimum percentage of years between 
+#'         start_yr & last_train_year that have data
 #'
 #' @return dataframe with original data and associated environmental data
 #' @importFrom dplyr "%>%" filter group_by summarise ungroup
-filter_ts <- function(bbs_data, start_yr, end_yr, min_num_yrs){
+filter_ts <- function(bbs_data, start_yr, end_yr, last_train_year, 
+                      min_year_percentage){
+  min_num_years = (last_train_year - start_yr + 1) * min_year_percentage / 100
+  browser()
```
ethanwhite: Remove debug line

R/forecast-bbs-core.R
```diff
@@ -224,27 +224,30 @@ get_env_data <- function(){
...
+#' all years between `start_yr` to `end_yr`.
 #'
-#' Attaches associated environmental data
+#' Also attaches associated environmental data
 #'
 #' @param start_yr num first year of time-series
 #' @param end_yr num last year of time-series
-#' @param min_num_yrs num minimum number of years of data between start_yr & end_yr
+#' @param last_train_year num last year of the time-series' training portion
+#' @param min_year_percentage num percent of years between of data between
```
ethanwhite: "years between of data between" -> "years with data between"?

davharris:
Thanks @ethanwhite. Fixed and merging now.

PR #128, opened 2017-06-16
Add cmip5

sdtaylor:
Adding downscaled cmip5 data for 2010-2050 bioclim variables. This needed a python script to do some initial cleaning of the data.


ethanwhite:
Very nice work. Thanks @sdtaylor!

PR #132, opened 2017-06-21
minor pipeline fixes

davharris:
* `AOU` -> `aou`

* simplified syntax for pulling out a vector from a data frame


sdtaylor:
Yeah, `pull` is pretty sweet.  

Changing all cases back from #120? Don't forget line 161.

davharris:
Ah, nice catch.

I think I know what's happening now.  Is it possible that the sqlite file you put on the T drive has an older version of the BBS data? Could you refresh the BBS-related tables to make sure? Sorry in advance if I'm wrong and this just makes extra work for you.

sdtaylor:
I just verified it with a fresh install of retriever 2.0, the headers are all lower case in the sqlite db and the csv version.

ethanwhite:
Looks like this is ready to go in now. Is that correct?

davharris:
@ethanwhite Now it should be. Sorry for the confusion on my end in #120, and thanks to @sdtaylor for catching what I missed in this one.

PR #136, opened 2017-07-06
2050

davharris:
This should allow everything but mistnet to predict to 2050.

### DIFF
--- a/R/post-observer.R
+++ b/R/post-observer.R
@@ -1,3 +1,27 @@
+iter_rep = function(df, .f, ...) {
+  # Get all distinct non-future iteration numbers
+  iters = df %>% 
+    ungroup() %>% 
+    filter(!is_future) %>% 
+    distinct(iteration) %>% 
+    pull(iteration)
+  
+  parallel::mclapply(
+    iters,
+    function(i){
+      # Run the function on the target iteration, retaining "future" rows for
+      # prediction. Finally, mark everything (including "future" predictions)
+      # with the iteration number.
+      df %>% 
+        filter(iteration == !!i | is_future) %>% 
+        .f(...) %>% 
+        mutate(iteration = !!i)
+    },
+    mc.cores = 8,
+    mc.preschedule = FALSE
+  )
+}
+
 #' @importFrom forecast Arima auto.arima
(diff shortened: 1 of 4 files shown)


R/post-observer.R
```diff
@@ -107,8 +137,8 @@ make_gbm_predictions = function(x, use_obs_model){
   
   cbind(test, mean = mean, model = "richness_gbm",  use_obs_model = use_obs_model, 
         stringsAsFactors = FALSE) %>% 
-    select(site_id, year, mean, richness, model, use_obs_model) %>% 
-    mutate(sd = sd)
+    select(site_id, year, mean, richness, model, use_obs_model, is_future) %>% 
+    mutate(sd = !!sd)
```
ethanwhite: What's the general use case for these double negations?

analysis/fit-predict.R
```diff
@@ -21,13 +24,22 @@ bioclim_to_discard = colnames(obs_model$data) %>%
...
+         is_future = FALSE) %>% 
   select(-ndvi_ann, -lat, -long, -site_index, -one_of(bioclim_to_discard))
 
 # Reclaim memory
 rm(obs_model)
 gc()
 
+future = get_env_data(timeframe = "future") %>% 
+  filter(site_id %in% !!x_richness$site_id,
+         year > !!settings$end_yr) %>%
```
ethanwhite: Based on my simple explorations of `!!` I'm confused by this. E.g.,

```
> !!3
[1] TRUE
> !!5
[1] TRUE
> 4 > !!3
[1] TRUE
> 4 > !!5
[1] TRUE
```

davharris:
🤦‍♂️ You're right. We should fit the model to 2013 for predicting the future. That will save a bunch of gymnastics like the `iter_rep` function. I'll close this issue and get a more straightforward version submitted today. Thanks for pointing this out.

`!!` is a new tidyverse thing for controlling non-standard evaluation (which is why it didn't do anything useful when you tried it with standard evaluation in your examples.  By default, `mutate(iris, f(Species))` applies a function to the column named `Species`.  If you want to refer to another object named `Species`, you can tell dplyr that using `!!`, and it looks in the parent environment instead of the data.frame.

PR #138, opened 2017-07-10
Full ms revision

ethanwhite:
This is a top to bottom revision of the ms. Major points of note:

* Completed expansion of Methods into prose
* Reworked flow of Results
* Recast @davharris's unnamed box item as 'Address unknown or unmeasured predictors'
* Added references throughout and created the bibtex file to go with them

### DIFF
--- a/manuscript/draft.Rmd
+++ b/manuscript/draft.Rmd
@@ -1,11 +1,13 @@
 ---
+title: Using best practices for forecasting biodiversity changes in breeding birds
+author:
+- David J. Harris
+- Shawn D. Taylor
+- Ethan P. White
 output:
   pdf_document: default
   html_document: default
 ---
-# Moving towards large-scale data-intensive forecasting of species richness in breeding birds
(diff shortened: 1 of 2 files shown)


ethanwhite:
GitHub is pretty terrible at diffs for this sort of work.

`git diff --word-diff master` will work much better locally.

I think it's also fine to just merge and then provide thoughts during your next revision on the entire ms.

davharris:
Thanks @ethanwhite! I had thought that there was a way to enable `word-diff` on Github with some URL magic, but turns out that's just for whitespace. I'll work on this once I've finished re-doing the future predictions code

davharris:
I think I'll merge and edit, like you suggested in your comment.

sdtaylor:
fyi I think github is doing a `--word-diff` when you press the "display the rich diff" button at the top of each file. It didn't work for Ethan's entire PR though, but it did work looking at individual commits.

PR #139, opened 2017-07-11
Revised crack at 2050 predictions

davharris:


davharris:
This should cover all the models except mistnet. Still need to test it for bugs and probably add some polish.

R/post-observer.R
```diff
@@ -105,10 +125,10 @@ make_gbm_predictions = function(x, use_obs_model){
     mean = mean + test$observer_effect
   }
   
-  cbind(test, mean = mean, model = "richness_gbm",  use_obs_model = use_obs_model, 
+  cbind(test, mean = mean, model = "richness_gbm", 
         stringsAsFactors = FALSE) %>% 
-    select(site_id, year, mean, richness, model, use_obs_model) %>% 
-    mutate(sd = sd)
+    select(site_id, year, mean, richness, model) %>% 
+    mutate(sd = !!sd, n.trees = !!n.trees, use_obs_model = !!use_obs_model)
```
ethanwhite: Thanks for the earlier explanation of `!!`. I understood (I think) the uses on lines 89 and 92, but I'm confused here. Where are the external `sd` and `n.trees` coming in. I don't see them passed to the function or in the above lines. If they are in `test` then I don't understand the use of `!!`.

R/sdm-rf.R
```diff
@@ -10,31 +11,34 @@ one_rf_tree = function(bbs, vars, sp_id, iter, use_obs_model, x_richness,
...
   vars = c(settings$vars, if (use_obs_model) {"observer_effect"})
   iters = sort(unique(x_richness$iteration))
   results = lapply(iters, one_rf_tree,
                    bbs = bbs, sp_id = sp_id, use_obs_model = use_obs_model, 
                    vars = vars,
                    x_richness = x_richness,
-                   last_train_year = settings$last_train_year) %>% 
+                   last_train_year = settings$last_train_year,
+                   future, future_observer_effects,
+                   settings) %>%
```
ethanwhite: Since you're using explicit names for all of the other variables I'd probably use them for future, future_observer_effects, and settings, for consistency.

R/sdm-rf.R
```diff
@@ -44,16 +48,18 @@ rf_predict_species = function(sp_id, bbs, settings, x_richness, use_obs_model){
...
+                               future, future_observer_effects) {
   
   out = parallel::mclapply(
     unique(bbs$species_id), 
     function(sp_id){
       rf_predict_species(sp_id, bbs = bbs, x_richness = x_richness, 
                          settings = settings, 
-                         use_obs_model = use_obs_model)
+                         use_obs_model = use_obs_model,
+                         future, future_observer_effects)
```
ethanwhite: Same idea as above on named arguments.

analysis/fit-predict.R
```diff
@@ -24,20 +34,49 @@ x_richness = obs_model$data %>%
...
+    select(which(colnames(.) %in% !!colnames(x_richness)))
+  
+  # Randomly-sampled observers for each `iteration`. Not used for the
+  # "average" model, which will just use observer_sigma directly without
+  # simulations.
+  future_observer_effects = obs_model$observer_sigma %>% 
+    purrr::map(~rnorm(nrow(future), 0, .x))
+} else {
+  future = NULL
+  future_observer_effects = NULL
```
ethanwhite: Missing indents for readability.

analysis/fit-predict.R
```diff
@@ -103,12 +144,12 @@ dir.create("rf_predictions", showWarnings = FALSE)
 
 gc() # Minimize memory detritus before forking
 
-rf_sdm_obs = rf_predict_richness(bbs = bbs, x_richness = x_richness, 
-                                 settings = settings, use_obs_model = TRUE,
-                                 mc.cores = 8) %>% 
-  saveRDS(file = "rf_predictions/all_TRUE.rds")
+rf_predict_richness(bbs = bbs, x_richness = x_richness, 
+                    settings = settings, use_obs_model = TRUE,
+                    future, future_observer_effects) %>%
```
ethanwhite: Same point about naming arguments (here and a few lines down).

R/observer_model.R
```diff
@@ -8,12 +8,11 @@
 fit_observer_model = function(stan_file = "mixed.stan", seed = 1,
                               output_file = "observer_model.rds", 
                               chains = 2, cores = 2, iter = 2000,
-                              thin = 4, ...){
+                              thin = 4, settings...){
```
davharris: Just a typo 😶

R/post-observer.R
davharris: I think that your understanding is correct, and they're both a few lines above my diff (see below). 

I could probably name `sd` better to avoid a name clash with the existing `sd` column. The `!!` for `n.trees` isn't necessary because there isn't (currently) a column with that name, but I like to be explicit.

https://github.com/davharris/bbs-forecasting/blob/b3344f6da1a50faec8647c85795ce1b2c1e9287d/R/post-observer.R#L118-L121

R/sdm-rf.R
davharris: 👍

ethanwhite:
That all looks sounds good. Pull the WIP and ping me when it's ready to go in.

davharris:
Removing "WIP"

Issue #133, opened 2017-06-26
NAs in "future" data


davharris:
I have NA for about 6% of the "future" bioclim data and about 2.5% of the NDVI data. Should I just throw those rows out, or is this a red flag we should investigate further?

sdtaylor:
How are you calculating those?  

I count 1193 unique sites in the past/prism data with bio1 as na, compared to 833 in the future/cmip data. Both sources have 5061 unique sites total.  
The past/prism data doesn't include any canada sites, while the future/cmip data does include canada data, so I would expect the past data to have more NA.

davharris:
I'm finding that there are 242 sites that had NAs in 2014 and 2015 (but had at least some non-NAs before). Each of these sites is missing all of its future bioclim data. I've listed them below.

I'm also finding a lot of rows in "future" data that have the year listed as 2051, and none of these have NDVI data.

Can you confirm?


```
c(2040L, 2041L, 2042L, 2070L, 2140L, 2141L, 2142L, 2217L, 11002L, 
11201L, 11210L, 11410L, 14004L, 14019L, 14038L, 14039L, 14045L, 
14050L, 14057L, 14075L, 14083L, 14099L, 14120L, 14123L, 14135L, 
14193L, 14345L, 18001L, 18006L, 18013L, 21001L, 21002L, 21006L, 
21007L, 21009L, 21010L, 21105L, 21107L, 25002L, 25007L, 25009L, 
25011L, 25016L, 25018L, 25025L, 25027L, 25032L, 25036L, 25039L, 
25051L, 25062L, 25064L, 25070L, 25075L, 25077L, 25078L, 25079L, 
25080L, 25083L, 25085L, 25086L, 25087L, 25088L, 25109L, 25116L, 
25132L, 25135L, 25170L, 25175L, 25178L, 25180L, 25183L, 25283L, 
27001L, 27056L, 27301L, 27305L, 35001L, 35101L, 35201L, 35301L, 
42001L, 42002L, 42004L, 42006L, 42008L, 42030L, 42031L, 42051L, 
42101L, 42102L, 42105L, 42108L, 42208L, 42311L, 42312L, 42411L, 
44002L, 44004L, 44005L, 44017L, 44023L, 44024L, 44025L, 44028L, 
44117L, 44308L, 46020L, 46027L, 46035L, 46036L, 46037L, 46041L, 
46042L, 46043L, 46048L, 46049L, 46058L, 46149L, 47001L, 47002L, 
[...]

PR #140, opened 2017-07-11
WIP: fix na issue

sdtaylor:
This caused a lot of unneeded NA's in the final cmip bioclim dataset. See #133


ethanwhite:
Thanks. The old NaN vs. NA weirdness. Love it.

PR #141, opened 2017-07-11
Revised methods and results

davharris:
Let me know if these changes all make sense.

### DIFF
--- a/manuscript/draft.Rmd
+++ b/manuscript/draft.Rmd
@@ -75,22 +75,13 @@ these approaches improve ecological forecasting in the future.
 
 ### Data
 
-Bird species richness was obtained from the Breeding Bird Survey (BBS)
-[@pardieck2017] using the Data Retriever software [@morris2013] and
+Bird species richness was obtained from the North American Breeding Bird Survey
+(BBS) [@pardieck2017] using the Data Retriever software [@morris2013] and
 rdataretriever R package [@mcglinn2017]. The BBS data was filtered to exclude
 all nocturnal, cepuscular, and aquatic species [since these species are not well
 sampled by BBS methods; @hurlbert2005], as well as unidentified species, and
-hybrids. Due to limitations of the environmental data (below), we did not
-include any data from outside the continental United States. Observed richness
-from 1982 (the first complete year of NDVI data) to 2003 were used for training
-the models, and data from 2004 to 2013 was used for testing their performance.
-Only BBS routes that were sampled during 70% of the years in the training period
-were used (i.e., routes with at least 16 annual observations). All data from
-routes that did not meet BBS acceptable criteria were excluded, as were routes
-that had not been visited during at least 70% of the years in our model-training
-window. The resulting dataset included [[N]] species and [[N]] sites. Among
-sites,  richness varied from [[N]] to [[N]] with an average richness of [[N]]
-species.
+hybrids. All data from routes that did not meet BBS acceptable criteria were
+also excluded.
 
 Environmental data included a combination of elevation, bioclimatic variables
 and a remotely sensed vegetation index (the normalized difference vegetation
@@ -109,6 +100,16 @@ SRTM 90m elevation dataset [@jarvis2008] obtained using the R package raster
 [@hijmans2016]. All environmental variables were aggregated to the average
 values within a 40 km radius (the length of a BBS route) of the starting point
 of each BBS route.
+
+We used observed richness values from 1982 (the first complete year of NDVI
+data) to 2003 for training the models, and data from 2004 to 2013 was used for
+testing their performance. We only used BBS routes from the continental United
+States (i.e. routes where PRISM data was available), and restricted the analysis
+to routes that were sampled during 70% of the years in the training period were
+used (i.e., routes with at least 16 annual observations). The resulting dataset
+included [[N]] transect runs of [[N]] unique sites, and included [[N]] species.
+Site-level richness varied from [[N]] to [[N]] with an average richness of [[N]]
+species.
 
 ### Accounting for observer effects
 
--- a/manuscript/draft.Rmd
+++ b/manuscript/draft.Rmd
@@ -105,8 +105,8 @@ We used observed richness values from 1982 (the first complete year of NDVI
 data) to 2003 for training the models, and data from 2004 to 2013 was used for
 testing their performance. We only used BBS routes from the continental United
 States (i.e. routes where PRISM data was available), and restricted the analysis
-to routes that were sampled during 70% of the years in the training period were
-used (i.e., routes with at least 16 annual observations). The resulting dataset
+to routes that were sampled during 70% of the years in the training period
+(i.e., routes with at least 16 annual observations). The resulting dataset
 included [[N]] transect runs of [[N]] unique sites, and included [[N]] species.
 Site-level richness varied from [[N]] to [[N]] with an average richness of [[N]]
 species.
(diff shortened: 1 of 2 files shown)


manuscript/draft.Rmd
```diff
@@ -78,25 +78,31 @@ these approaches improve ecological forecasting in the future.
...
+from 1982 (the first complete year of NDVI data) to 2003 were used for training
+the models, and data from 2004 to 2013 was used for testing their performance.
+Only BBS routes that were sampled during 70% of the years in the training period
+were used (i.e., routes with at least 16 annual observations). All data from
+routes that did not meet BBS acceptable criteria were excluded, as were routes
+that had not been visited during at least 70% of the years in our model-training
+window. The resulting dataset included [[N]] species and [[N]] sites. Among
+sites,  richness varied from [[N]] to [[N]] with an average richness of [[N]]
+species. Within-site richness variation (among years) was small (standard
+deviation of about 5 species, after controlling for observer effects).
```
ethanwhite: I think I prefer the original separation of the discussion of the data itself and then later the decisions we made with respect to modeling it. This feels a bit like two things alternatingly mixed together. However I don't feel that strongly about it.

Either way I think we should cut the last sentence here and move it to the results if it isn't there already.

manuscript/draft.Rmd
```diff
@@ -259,59 +260,63 @@ on these time scales once observer differences are accounted for.
...
-for forecasting within site variability was either the "average" baseline model
-or the ARIMA model (Figure 3). In general, each model's error increased
-more-or-less linearly with time, although the slopes and initial error rates
-varied across models (Figure 3). The site-level models (i.e. the baselines and
-the ARIMA methods) tended to do very well (Figure 3) on short time horizons,
-especially in terms of RMSE.
+for forecasting within-site variability was either the "average" baseline model
+or the ARIMA model [[Not sure what the bit about within-site variability means
+here. Also, didn't Naive usually win when observer effects weren't corrected
+for?]] (Figure 3). In general, each model's error increased more-or-less
```
ethanwhite: The "within-site variability" part is meant to distinguish this analysis from that in the previous paragraph on among site comparisons.

Yep, the Naive is usually the winner when not correcting for observer effects, but my thinking was that our main results are those when we've corrected for observer effects. We want to discuss those changes in the context of the importance of the observer model, but I don't think we want to mix that discussion in with the other results. That's why I held discussion of the non-observer corrected data out for it's own paragraph. Does that make sense?

manuscript/draft.Rmd
```diff
@@ -259,59 +260,63 @@ on these time scales once observer differences are accounted for.
...
-are generally far too confident, with observed richness values frequently
-falling well outside their confidence limits (Figure 3C). The "naive" model was
-the only model whose confidence intervals were too wide (Figure 3), which can be
-attributed to the rapid rate at which these intervals expand as the time lag
-increases (Figure 1). The JSDM and macroecological model has the best coverage.
+stack was the most overconfident, with observed richness values frequently
+falling well outside their confidence limits (Figure 3C). As noted elsewhere
+[@harris2015], this overconfidence is a product of the assumption in stacked
+SDMs that errors in the species-level predictions are independent. The JSDM and
+macroecological model has the best coverage, though each model still
```
ethanwhite: "has" -> "had" (while acknowledging that it was my typo that you just copied and pasted.)

manuscript/draft.Rmd
```diff
@@ -259,59 +260,63 @@ on these time scales once observer differences are accounted for.
...
-latent variables were already accounting for observer-related uncertainty. In
-combination changes in point estimates and coverage resulted in changes in the
-best models based on deviance at different time scales. The macroecological
-model and the JSDM appeared to narrowly outperform the average and ARIMA models
-when the forecast horizon was greater than five years when not correcting for
-observers, but after the correction the average and ARIMA models maintained a
-small average advantage out to at least ten years.
+latent variables were already accounting for observer-related uncertainty. These
+differences meant that, in terms of deviance, neither the JSDM nor the
+macroecological model model benefitted as much from the as the site-level models
```
ethanwhite: delete extra "model and "from the"

manuscript/draft.Rmd
davharris: Makes sense.  I'll see if I can introduce subsections like we had in the Methods section to avoid this sort of confusion from readers.

davharris:
I think this ordering of the Results makes it easier to follow. The discussion of uncertainty and deviance is more integrated now, in particular. Very open to criticism on this commit, though.

Still figuring out how best to arrange the Data section, apart from moving the SD information to the Results section.

davharris:
Okay, took a shot at the Data section.  Ready for another review when it's convenient.

manuscript/draft.Rmd
```diff
@@ -253,71 +252,60 @@ The site-observer model found that 70% of the variance in richness in the
...
+sites resulting in a notable clustering of predicted values near ~60 species and
+the poorest performance in the cross-site comparison (R^2 XX-XX%).
+
+In general, each model's error increased more-or-less linearly with time,
+although the slopes and initial error rates varied across models (Figure 3). The
+site-level models (i.e. the baselines and the ARIMA methods) tended to do very
+well (Figure 3), especially on short time horizons, and especially in terms of
+RMSE. Of these, the "Naive" model performed especially well at sites with
+uncorrected observer turnover (since its predictions were best-calibrated to
+observers that were active in later years).
```
ethanwhite: Two thoughts about the revision of this paragarph:

1. I still don't personally see the value of focusing on the performance of the uncorrected model when discussing general overall performance. Can you walk me through you thinking on why it's useful to include here rather than saving those results for the paragraph on the observer model? Also, since we decided not to include non-observer model results in Figures 3 and 4 we'd need to jump to citing 5 here to bring this up at this point. That's OK if we need to, but I think that's part of my general confusion over the value of talking about this here.

2. Since we just finished a paragraph talking generally (i.e., not with respect to time lag) about the among sites comparison it seems weird to then move to the within site dynamics and then not talk about them generally before delving into the changes with time horizon. That's why I added the opening sentence speaking generally about the results, to help with the transition and provide a high level summary of the results.

If we end up adding some version of the current opening sentence back in I think we should probably plan to flip the order on Figures 3 and 4, so that the violin plots come first and get cited in the first sentence and the temporal trends come second.

manuscript/draft.Rmd
```diff
@@ -253,71 +252,60 @@ The site-observer model found that 70% of the variance in richness in the
...
+
+Most models produced confidence intervals that were too narrow, indicating
+overconfident predictions. The random forest-based SDM stack was the most
+overconfident model, with observed richness values frequently falling well
+outside its confidence limits (Figure 3C). This stacked SDM's narrow
```
ethanwhite: "This" -> "The

manuscript/draft.Rmd
```diff
@@ -253,71 +252,60 @@ The site-observer model found that 70% of the variance in richness in the
...
+predictive distribution also caused it to have notably higher deviance (Figure
+3B) than the next-worst model, even though its point estimates were not
+unusually bad in terms of RMSE (3A), and the other models' relative performance
+was not nearly as sensitive to the choice of error metric. As discussed
```
ethanwhite: I'm not sure that "was not nearly as sensitive to the choice of error metric" adds a lot here and I think it's arguable with respect to mistnet which goes from consistently being the worst wrt RMSE, has the best coverage, and is in the pack for deviance.  Having that sentence also breaks up the flow on the route to the next sentence, which is also focused on the stacked SDMs.

manuscript/draft.Rmd
```diff
@@ -253,71 +252,60 @@ The site-observer model found that 70% of the variance in richness in the
...
+elsewhere [@harris2015], this overconfidence is a product of the assumption in
+stacked SDMs that errors in the species-level predictions are independent.
+Because they avoided this assumption, the GBM-based "macroecological" model and
+especially the mistnet JSDM had reasonably well-calibrated uncertainty
+estimates, even in the presence of uncorrected observer biases (Figure 3C). This
```
ethanwhite: Again, not sure why we want to talk about the uncorrected observer biases here instead of in its own paragraph.

manuscript/draft.Rmd
```diff
@@ -253,71 +252,60 @@ The site-observer model found that 70% of the variance in richness in the
...
-did from the observer corrections. As a result, the average and ARIMA models
-maintained a small average advantage over the JSDM in terms of deviance at
-longer time scales, rather than falling behind after about five years.
-
+resulted in improvements to nearly all metrics of fit (Figure 4). The
+improvements in point estimates (as measured by RMSE) primarily resulted from
+large improvements in a small number of forecasts where a large difference
+between observers resulted in a large shift in predicted richness values that
+the models could not otherwise accommodate. The "Average" and ARIMA models
+generally showed the largest improvement here, and the Naive model showed the
```
ethanwhite: delete "here"

PR #143, opened 2017-07-12
Minor discussion edits

sdtaylor:
@ethanwhite I think all the discussion edits you did look good. I'd like to look at all the statements one more time when we have figures to look at. This is just fixing a few typos and adding in some more citation info.


davharris:
Looks good to me. Thanks for the tip about rich diffs. I'd tried it on multi-commit diffs and assumed it wasn't relevant.

PR #147, opened 2017-07-13
Add long term methods

sdtaylor:
I think this paragraph might fit here.   

The supplement table is straight from the downscaled cmip website. I only included models that we used. Tredennick 2016 used this same table so I think it should be fine.


ethanwhite:
LGTM. Will leave it to @davharris to merge once he's had a look.

davharris:
Looks good to me, except that the term "naive" is already being used to describe a forecasting method. Maybe "simple"?

sdtaylor:
What about "naive NDVI forecast"? We'll run out of adjectives pretty quick
if we can only use naive once.

davharris:
I think we have plenty of adjectives available (see below). We're currently using "naive" to basically mean that we project forward based on 2013 alone, in direct contrast to averaging over all the years. So using "naive" to mean "average" is confusing, in my opinion.

In approximately descending order of preference, here are some alternatives I've come up with: 

* simple
* simplified
* basic
* rough
* primitive
* bare-bones
* vanilla
* lightweight
* makeshift

sdtaylor:
Ok, simple it is

PR #161, opened 2017-08-15
Big manuscript edits

davharris:
Also some changes to the code for "future" data and plotting.

Also a little R script (`manuscript/knit.R`) that deals with the cases where R code got printed in the PDF instead of evaluated.

Still need to finish polishing/labeling the figures and adding some citations, but otherwise I think this is getting really close.


ethanwhite:
Overall this looks great. And I'd be happy to merge it and then move on to the next round of editing with the exception of the complete removal of the Discussion paragraph on how error increases over time. I think it's a pretty important paragraph both from a conceptual perspective (these kinds of issues are likely real and it would be good to talk about them here) and because a natural point of skepticism about our study is the relatively short time scale of our hindcast test period, which this paragraph addresses directly.

davharris:
@ethanwhite That paragraph wasn't accurate anymore according to the results I had. I didn't see any sign that mistnet and GBM were catching up anymore.  It's possible that the paragraph will become accurate again once we fix the 1982 data though. I thought that we were still talking about increasing error over time elsewhere, but it's possible I'm mistaken.

Maybe the best option is to leave that paragraph in for now, with a [[note]] that we should double-check it after the final models are run?

ethanwhite:
> Maybe the best option is to leave that paragraph in for now, with a [[note]] that we should double-check it after the final models are run?

Yeah, that sounds good. I think we need the time-horizon/scale discussion regardless of whether we see the convergence or not, so it might need to be edited, but not removed.

ethanwhite:
Also, is this based on the current results on the T-drive?

davharris:
> it might need to be edited, but not removed.

I'll add the paragraph back in, with a note

> Also, is this based on the current results on the T-drive?

Yes

ethanwhite:
Thanks @davharris, in it goes.

PR #169, opened 2017-08-18
Editorial changes to abstract

ethanwhite:


ethanwhite:
I'm going to PR these edits in section by section to give us more granularity for discussion, further edits, and improvements, with the goal of having everything that goes in be something that everyone is happy with as a near final form.

PR #173, opened 2017-08-18
Editorial work on Methods section

ethanwhite:
This is built on 9ec2b2b since Table 1 was moved into the methods from the intro.


ethanwhite:
Replaced by #174 (rebased version)

PR #175, opened 2017-08-19
plotting code

davharris:
This fixes a bunch of stuff, including labels and colors. But I don't have the mistnet results on this machine, so mistnet is missing from all of the PNGs. 🙄

Will fix soon. In the mean time, feel free to let me know if you have any thoughts on the style/formatting/labeling/etc.


ethanwhite:
Flipping B and C in the performance over time figure would probably match more naturally with the way things are written up in the ms.

davharris:
Okay, cool. I've added mistnet, cleaned up some labels, and flipped panels B and C like @ethanwhite  suggested. Ready for review now.

ethanwhite:
Why are the densities such much smaller in the bottom panel of `observers.png` than the top?

davharris:
> Why are the densities such much smaller in the bottom panel of observers.png than the top?

This is tricky.  The Naive Baseline's Deviance violin is made almost entirely of zeros, so it's approaching a strip with zero height (and thus zero area; see below). Within a panel, the violins all have the same area, so the other ones are tiny too.

image

We have a few options. In no particular order:

1) Switch to boxplots or histograms or something else.

2) Re-scale everything so that the violins have constant width instead of constant area.  This doesn't really solve the problem, since some violins will still end up much larger than others. Arguably, this option is worse, since the differences in visual weight will occur within panels (instead of between them, as is happening now).

3) Tweak the bandwidth parameters for the violins.  This is probably our best option if we can make it work, so I'll start playing with it.

ethanwhite:
Thanks for the explanation. I don't think this is a high priority since we could submit it this way and clean it up later if necessary.

ethanwhite:
Other than the oddity with the observer model violins this all looks great. In it goes!

PR #177, opened 2017-08-23
Minor wording suggestions

davharris:
Tweaks of Ethan's wording in 2 places


ethanwhite:
Should it be "forecasted" or "forecast"... how about "Given forecasts for..."

Other than that, LGTM.

davharris:
I like "Given forecasts for..." a lot.  Will update and merge.

PR #182, opened 2017-08-24
Shawns final final edits

sdtaylor:


ethanwhite:
Very nice work @sdtaylor.

@davharris - these all seem clear enough to me that I'm going to go ahead and merge them to work on top of these changes, but if you see anything in the PR you don't like just let me know and I can change it back in my edits.

PR #183, opened 2017-08-25
Edits to the Results section

ethanwhite:


ethanwhite:
Note to be careful using rich diff for this since the figure legends don't display in rich diff.

davharris:
This looks good to me, especially the figure captions. Thanks!

PR #192, opened 2017-09-01
Add obs_arrows

davharris:
Note that this changes the value of our `d` data.frame (it now has both observer and non-observer results).

(I also changed the y-axis limits on the performance-time plot slightly for boring technical reasons that are no longer relevant)


ethanwhite:
LGTM.

PR #193, opened 2017-09-03
Early discussion edits

davharris:
* Weaken/re-focus caveats at the end of the first discussion paragraph

* Try to strengthen the discussion of SDMs and uncertainty in the next two paragraphs

(Also fix a typo in the abstract)


ethanwhite:
Really nice work on the writing!

My one question is about removing the birds in North America limitation. This definitely going to get brought up by others. It's pretty common to see this critique on reviews of anything large scale these days and it is a central argument against empirical results showing minimal trends in local scale richness (i.e., they are present, we're just not looking at the places where they are happening, e.g., the tropics).

I see a couple of options.

1. We could move this language into the methods. It would be less likely to mess with flow there, and it definitely did mess with the flow in the Discussion the way I'd written it.
2. Modify the last sentence to something like "While this study is restricted to breeding birds in North America, its results are consistent with a growing literature on the limits of ecological forecasting, as discussed below." It's almost as smooth and keeps the explicit details of the limitation.

davharris:
Option 2 works well for me. Fixed.

PR #196, opened 2017-09-03
update figure captions

davharris:
(note that figure caption edits don't show up in the rich diff)

### DIFF
--- a/manuscript/draft.Rmd
+++ b/manuscript/draft.Rmd
@@ -214,12 +214,13 @@ downstream models with each of the Monte Carlo samples. Each Monte Carlo sample
 represented a different possible set of observer-level and site-level random
 effect values across the full 32-year dataset.
 
-![**A.** Model predictions for route 72035 when all observers are treated the
-same (black points). **B.** Model predictions for the same route when accounting
-for systematic differences between observers (represented by the points'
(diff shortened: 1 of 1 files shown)


manuscript/draft.Rmd
```diff
@@ -214,13 +214,12 @@ downstream models with each of the Monte Carlo samples. Each Monte Carlo sample
...
 effect values across the full 32-year dataset.
 
-![**A.** Model predictions for [[STATE]] route [[ROUTE NUMBER]] when all
-observers are treated the same (black points). **B.** Model predictions for the
-same route when accounting for systematic differences between observers
-(represented by the points' colors). In this example most models are made more
-robust to observer turnover. Note that the "naive" model is less sensitive to
-observer turnover, and does not benefit as much from modeling
-it.](../figures/observer_predictions.png)
+![**A.** Model predictions for route 72035 when all observers are treated the
```
ethanwhite: Since we create these route numbers we should break them out into the way BBS describes things. In this case that would be "Model predictions for Pennsylvania route 35...". (state numbers: https://www.pwrc.usgs.gov/bbs/state_codes.htm)

davharris:
Good catch. I forgot we made those route numbers ourselves.  Merging now.

PR #198, opened 2017-09-05
Discussion edits

davharris:
The biggest change here is for the way lead times are discussed. We don't really have evidence that the models' performance was converging, so I cut a lot of material there and tried to re-organize it in a bigger-picture way.

### DIFF
--- a/manuscript/draft.Rmd
+++ b/manuscript/draft.Rmd
@@ -594,9 +594,12 @@ this study show little change in richness [site-level SD of `r numbers$resid_sd`
 species, after controlling for differences among observers; see also
 @lasorte2005]. The absence of rapid change in this dataset is beneficial for the
 absolute accuracy of forecasts across different sites: when a past year's
-richness is already known, it is easy to estimate future richness.  The
-site-level stability of the BBS data also explains why SDMs and macroecological
-models perform relatively well at predicting future richness, despite failing to
+richness is already known, it is easy to estimate future richness. Ward et al.
+(2014) found similar patterns in time series of fisheries stocks, where
+relatively stable time series were best predicted by simple models and more
+complex models were only beneficial with dynamic time series. The site-level
+stability of the BBS data also explains why SDMs and macroecological models
+perform relatively well at predicting future richness, despite failing to
 capture changes in richness over time. However, this stability also makes it
 difficult to improve forecasts relative to simple baselines, since those
 baselines are already close to representing what is actually occurring in the
--- a/manuscript/draft.Rmd
+++ b/manuscript/draft.Rmd
@@ -562,22 +562,20 @@ the lead times relevant for conservation and management. Two studies have
 assessed models of species richness at longer lead times [@algar2009;
 @distler2015], but the results were not compared to baseline or time-series
 models (in part due to data limitations) making them difficult to compare to our
-results directly. Given the potential changes with lead time of the models and
-processes are most relevant to forecasting, it is important that more forecasts
-with longer lead times are evaluated and that the inferences drawn from
-short-term and long-term forecasting studies be compared. By evaluating
-shorter-term forecasts, researchers can avoid waiting so many decades to
-evaluate their methods [@petchey2015; @tredennick2016; @dietze2016]. However,
-since drivers of species richness differ at different temporal scales
-[@rosenzweig1995; @white2004; @white2007], short-term evaluations should be
-viewed as informing, not replacing, long-term evaluations [@tredennick2016]. In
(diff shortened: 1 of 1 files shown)


ethanwhite:
I guess I misunderstood where our conversation on this had ended up. My impression was that based on our agreement with @sdtaylor's last statement that

> different models still have different rates of error increase though, so the idea of convergence is still sound and can be discussed.

that we were all OK discussing this as a possibility.

ethanwhite:
> Given the potential changes with lead time of the models and processes are most relevant to forecasting

Currently comes out of nowhere.

manuscript/draft.Rmd
```diff
@@ -640,42 +633,37 @@ the habitats and species being surveyed [@sauer1994]. Accounting for differences
...
-error can be important for properly estimating and reducing uncertainty in
-forecasts and can also lead to changes in the best methods for forecasting [Box
-1]. Another key aspect of an ideal observation model is imperfect detection. In
-this study, we did not address differences in detection probability across
-species and sites [@boulinier1998] since there is no clear way to address this
-issue using NA Breeding Bird Survey data without making strong assumptions about
-the data [i.e., assuming there is no biological variation in stops along a
-route; @white2010], but this would be a valuable addition to future forecasting
-models.
+improving most models' point estimates relative to the naive baseline [Box 1].
```
ethanwhite: This Box 1 citation looks like it is associated with "improving most models' point estimates relative to the naive baseline", which doesn't make sense to me.

manuscript/draft.Rmd
```diff
@@ -640,42 +633,37 @@ the habitats and species being surveyed [@sauer1994]. Accounting for differences
...
 effects resulted in changes in which models performed best, most notably
-reducing the relative performance of the naive model's point estimates. This
-suggests that, prior to accounting for observer effects, the naive model
-performed well because it was capable of accommodating rapid shifts in estimated
-richness introduced by changes in the observer. These kinds of rapid changes
-were difficult for the other single-site models to accommodate, and so the
-average performance of the ARIMA and average models improved once this source of
-observation error was addressed. This demonstrates that modeling observation
-error can be important for properly estimating and reducing uncertainty in
-forecasts and can also lead to changes in the best methods for forecasting [Box
```
ethanwhite: I like this sentence in that it ties back in to the overarching theme of the ms and reminds readers that this particular method is important to the overall goals of better forecasting. What's your thinking on the reason to take it out?

sdtaylor:
The specific models, JSDM GBM and the average, are probably not converging. But there is still different rates of decay among all models, so perhaps this can be more generic. 

The accuracy for forecasts is expected to decline as forecasts are made further into the future, so it is 
important to evaluate forecasts at different lead times [Box 1; @petchey2015]. _For deviance (the error metric that evaluates both models' point estimates and their uncertainty simultaneously) there is different rates of decay among some models, with the stacked SDM decaying the fastest followed by the naive model. The remaining models all have similar rates of error decay. Depending on which models are evaluated this leads to the potential convergence of models over time, where models which perform worse at short lead times then perform better at longer lead times._ This potential convergence is biologically plausible because at short lead times, environmental changes are relatively small and changes in the biota may lag behind changes in the environment. However, as the lead time of the forecast increases the amount of environmental change should increase and the systems will have more time to respond, potentially resulting in shifts towards improved relative performance of models incorporating environmental and ecological information [@blonder2017]. This is potentially important for interpreting our results...

davharris:
How's this for the paragraph we discussed? https://github.com/weecology/bbs-forecasting/pull/198/commits/3760770078aaca29c7fdbcb9e5d74f4c6508de0c

I think it might need one more sentence at the end, but it flows better and should act as a better caveat/appeasement mechanism.

sdtaylor:
The blonder2017 ref doesn't make sense without the discussion about time lags in environmental responses. But it can probably just go with the group of refs just prior to it.

sdtaylor:
Otherwise I like it. I did the stats on the trends as well and agree with the last statement. 

image

davharris:
Great. I've moved the Blonder citation and will deal with Ethan's other two comments next.

manuscript/draft.Rmd
davharris: I honestly don't know why removed it. Putting it back in now..

davharris:
Thanks guys! Ready for another review.

ethanwhite:
Nice work @davharris! This is good for me. @sdtaylor if you're happy go ahead and merge.

sdtaylor:
:+1:

PR #199, opened 2017-09-05
variance partitioning

### LINKED ISSUE #187: Interpretation of Figure 6
Currently Figure 6 is cited for:

> the majority of the error comes from errors in the site-level mean

and also

>  none of the methods evaluated reliably determined whether site-level richness would increase or decrease over time [in contrast to: When comparing forecasts across sites, many different approaches to forecasting produce reasonable forecasts]

These seem contradictory from a first glance.

We either need to improve the language or pull this figure. We only really use it in the first paragraph of the discussion where a citation to Figure 4 would work equally well.

davharris:
Fix #187, maybe


ethanwhite:
I think I'm still struggling with what the key take home point is. I thought I understood it when you made the figure originally, but I'm struggling now in context. Can you give us a one sentence summary of what this figure tells us?

davharris:
Main point:

> Environmental models have larger biases at the site level than single-site models [because they can't distinguish among different sites with similar climates/NDNVI/elevation]; they also don't succeed in tracking changes in richness over time [i.e., their total error would be lower if they just averaged their predictions over the 10 years instead of trying to predict how things change]

Secondary point, which might be getting too much attention in the current draft and causing some of the confusion:

> Even though all the models do pretty well in terms of reducing the site-to-site variance (from 140-ish to 20-ish), that error component is still where we find the bulk of the residual error.

ethanwhite:
OK, thanks. So the primary comparisons are between models not between site-level and annual within a model. In that case what do you think about splitting the figure based on error component rather than model type? I'm not attached to the change, but I think my confusion (which may just be me) is that the way it's currently done my eye goes straight to "annual < site-level" instead of the central point which is "time-series < environment".

davharris:
How about this? 

image

ethanwhite:
Yeah, that looks great.

PR #200, opened 2017-09-05
add ward et al. to refs

davharris:
I assume this is the correct paper?


ethanwhite:
Yep, that's the one.

davharris:
Okay, great. Merging.

PR #203, opened 2017-09-11
Minor pre-submission changes

davharris:
* Citations
* Don't promise non-existent appendix


davharris:
Can we merge this? I think (but haven't confirmed) that it was included in the version I submitted to PeerJ.

PR #204, opened 2017-11-06
clarify hindcasting to reviewer

sdtaylor:
do you think this will be enough for this comment? or should I add clarification in the manuscript?


ethanwhite:
I think we might as well add a *sensu* and a couple of citations to the ms just to cover our bases.

PR #209, opened 2017-12-04
Reviewer comments: spatial autocorrelation

davharris:
I spent a long time trying to get this right. I'm still not sure it's optimal, but I thought I'd see what you two thought.

### DIFF
--- a/manuscript/reviewer_reply.Rmd
+++ b/manuscript/reviewer_reply.Rmd
@@ -57,6 +57,8 @@
 
 *Line 140: I’m curious why you didn’t consider spatial structure in the random effects here, using GMRF models or similar? One hand these models might be overkill if the habitat/environmental variables (which are spatially correlated) explain variability in richness, but the spatial correlation may pick up other sources of variability not accounted for by those variables.*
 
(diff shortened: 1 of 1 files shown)


sdtaylor:
So you think the environmental models won't benefit much from including spatial autorcorrelation because the bioclim vars have a similar correlation and thus the models already pick-up much of that?

ethanwhite:
I guess I'm still struggling with why we would want to do this. The goal of that model is to do one thing, control for observer effects. Since observers typically survey the same site we can't just include observers because we'd end up with a lot of site effects being designated as observer effects, so we put the site in to make sure that we didn't end up correcting out real differences between sites. Unless I'm missing something I'd just say something like:

> The goal of this model is to control for observer effects, not to provide a focal model of the system (though we do use it for the average model as a convenience). Because observers typically observe the same site year after year it was necessary to include site in this model to avoid classifying variation among sites as differences among observers and incorrectly removing it before fitting other models. Given the use of this model it is unclear what including spatial structure would accomplish and it seems possible that it could end up correcting out meaningful ecological patterns and therefore confuse the comparison of more complex models to the baselines.

davharris:
Thanks @ethanwhite, that's definitely better than my version. I think your last sentence is especially good. I was worried about (but did *not* articulate well) the possibility that the observer effects could be corrupted by spatial patterns, as we talked about before. But that's probably too subtle to deal with here. I'll probably replace my paragraph with your wording or something very similar.

Thanks again.

davharris:
I've posted a slightly edited version of Ethan's comments; see commit message in f7b1dd3. Thanks again @ethanwhite for helping focus this response in a better direction!

ethanwhite:
LGTM

sdtaylor:
:+1:

PR #211, opened 2017-12-06
Fix figures

davharris:
Most of the changes are visually indistinguishable. Apparently I used a different random seed for the single-species SDMs when I re-ran everything, so a few of those results jiggled around a small amount.

I had forgotten to push the addition of "A" and "B" to the code for the observer figure, so that's now added.


sdtaylor:
this is where the swipe feature really shines

sdtaylor:
They asked for a color legend for Fig. 2. I think it's fine as is. I also got a lab mate opinion and they thought the colors were fine as well. But it was suggested to improve this sentence in the Fig 2 legend. 

In this example most models are made more robust to observer turnover **by including an observer model**.

davharris:
Thansk @sdtaylor. Like this? https://github.com/weecology/bbs-forecasting/pull/211/commits/4dfdedf2cadb1c91d69be47534b6b5fffa7e1d0d

ethanwhite:
LGTM.

sdtaylor:
:+1:

PR #212, opened 2017-12-06
Equations

davharris:
(Also fixed a typo and the right-justification)


sdtaylor:
this looks great, I really like including the stand code along with the equations

davharris:
Thanks Shawn! Merging

PR #213, opened 2017-12-06
misc updates/responses

davharris:
Mostly minor, except where I tried to clarify what the mixed model was doing.


sdtaylor:
Wouldn't all 3 environmental models be robust to uninformative covariates? If BRT are considered robust than random forests should be as well, and since it's based on a neural net I would guess MistNet is as well.

ethanwhite:
It does read a little oddly to call out a single method as being robust and not include the others. If describing all three this way is justified then I like @sdtaylor's suggestion of mentioning them all. If not I think it's OK to say that this is complicated by the model fitting procedure and (if we all agree that what follows is correct) that variable selection is typically less important for prediction than for inference.

The reality is that we've done a massive amount of comparative modeling here and reviewers will always be interested in seeing you go one step further than you have. We've already made a lot of big steps forward in this paper and it's fine to stop there.

davharris:
I had thought he was just asking about the "macroecological model" based on the line numbers, but maybe I misinterpreted.  Will take another stab at this soon.

davharris:
You two were right and I misread: he did ask about all environmental models, not just boosting. Fixed now.

sdtaylor:
:+1: LGTM

PR #215, opened 2017-12-07
response to cmip5 question

sdtaylor:


davharris:
Thanks Shawn! Merging

PR #219, opened 2017-12-08
more results in the abstract?

davharris:
Reviewer 2 wanted more results in the abstract. I think that this addition is too long and breaks the flow, but a revised version might improve the paper.


sdtaylor:
How about: 


... failing to contain the true value at 4-6 times the nominal rate. These opportunities for improvement in biodiversity forecasting, as well as the forecasts themselves and the relative performance of different modelling approaches were directly influenced by the best practices implemented in this study. To facilitate the rapid improvement of biodiversity forecasts, we emphasize the value of specific best practices in making forecasts and evaluating forecasting methods.

ethanwhite:
See my attempt at this in #222.

PR #221, opened 2017-12-08
edits regarding "no directional change"

davharris:
Would like to hear your thoughts on this.


ethanwhite:
Did something go wrong here. The first edit now results in a sentence:

"While this study is restricted to breeding birds in simple baselines."

ethanwhite:
Chatted with @davharris about this on Slack. I'm working on this now.

ethanwhite:
I think this material is better placed near the two paragraphs that we already have discussing this to avoid breaking up the discussion of it. I've tried an alternative approach in #223 that splits off a bit of our 2nd paragraph on this topic to create a new paragraph addressing two of Review 2's comments as well as the AEs. See what you think.

sdtaylor:
This sentence " we expect that long-term SDM-based forecasts will generally be at least as poor as decade-scale ones." doesn't agree with what we already said elsewhere: 

 In general, drivers of species richness can differ at different
 temporal scales (Rosenzweig 1995, White 2004, 2007, Blonder et al. 2017), so different
methods may perform better for different lead times. In particular, we might expect
environmental and ecological information to become more important at longer time
scales, and thus for the performance of simple baseline forecasts to degrade faster than
forecasts from SDMs and other similar models.

davharris:
Good catch @sdtaylor. Closing this.

manuscript/draft.Rmd
```diff
@@ -577,8 +584,8 @@ they also show that stacked SDMs dramatically underestimate the range of
 possible outcomes; after ten years, more than a third of the observed richness
 values fell outside the stacked SDMs' 95% prediction intervals. Consistent with
 @harris2015 and @warton2015, we found that JSDMs' wider prediction intervals
-enabled them to avoid this problem. Macroecological models appear to share this
-advantage, while being considerably easier to implement. 
+mostly enabled them to avoid this problem. Macroecological models appear to share
```
sdtaylor: @davharris did you want to keep this "mostly" comment?

PR #222, opened 2017-12-08
Add more specific results to the Abstract

ethanwhite:
An alternative to #219, which focuses on giving specific examples of where best practices were useful.


sdtaylor:
I like this one a little better than #219, it gives more takeaways while also emphasizing the box bullet points in each one.

davharris:
Cool. I'll close #219 and merge this. Thanks both.

PR #223, opened 2017-12-08
Revise Discussion in response to questions about forecastability

ethanwhite:
This addresses the concerns of both the associate editor and Reviewer 2 about
the forecastability of BBS richness in the chosen time period. In adds a new
paragraph to the Discussion that expands on our previous discussion of these
points and calls for future research in this area.

Includes associated updates to the letter.

Alternative approach to #221.

### DIFF
--- a/manuscript/draft.Rmd
+++ b/manuscript/draft.Rmd
@@ -598,7 +598,7 @@ different temporal scales [@rosenzweig1995; @white2004; @white2007;
 In particular, we might expect environmental and ecological information to
 become more important at longer time scales, and thus for the performance of
 simple baseline forecasts to degrade faster than forecasts from SDMs and other
-similar models.  We did observe a small trend in this direction: deviance for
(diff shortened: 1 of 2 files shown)


davharris:
Thanks Ethan. I think your changes to the text are better than mine.  

Looks good to me, except "most ecological forecasts are intended to be applied to the immediate future" might not be quite right.  Most *published* SDM predictions are for 2050 or some far-off date, I think. Although they're not really "intended to be applied" to anything.

ethanwhite:
That's a good point about that language. What I was trying to convey was "we want to predict into the future where we expect this 'incredible dynamism' to continue, not to some stable point that occurred in the past."

ethanwhite:
OK, I just tweaked the language to try to same more directly what I was thinking. See what you think.

sdtaylor:
LGTM

davharris:
LGTM. Thanks Ethan!

PR #227, opened 2017-12-09
answer about the long term forecasts

sdtaylor:
I think this is discussed pretty well in the text so I didn't change anything there.


davharris:
An alternative: https://github.com/weecology/bbs-forecasting/commit/c1a3830eeee66dcd71c30905fc05fff4803fd179?short_path=2112e0d#diff-2112e0db2b5026e7808dfd94ee62b151

(sorry for posting this in the wrong thread initially)

sdtaylor:
@davharris version is better

PR #229, opened 2017-12-09
Add response describing binder as solution to vignette request

ethanwhite:
Addresses Ward's "Line 266" comment by stating our plan to enable binder instead of adding a vignette.
I want to do this anyway and I think it's the better solution. Since the R support is still in development we'll
have to do this after we resubmit.


davharris:
LGTM

sdtaylor:
:+1:

PR #230, opened 2017-12-09
Reword sentence that used "forecast" three times in 13 words

ethanwhite:
Addresses Ward's comment about Line 329.


davharris:
LGTM

sdtaylor:
can you add a comment to the reply letter?

ethanwhite:
> can you add a comment to the reply letter?

Done.

PR #232, opened 2017-12-09
equations for baselines

davharris:
If we like this, I can add a comment for the reviewers.  

Ideally, we should also discuss where the standard deviations of ε come from, but that differs for the observer-corrected and non-corrected "average" models so I can't figure out the best way to discuss it without breaking the flow.

### DIFF
--- a/manuscript/draft.Rmd
+++ b/manuscript/draft.Rmd
@@ -298,23 +298,29 @@ noise around a site-level constant:
 
 $$y_t = \mu + \epsilon_t.$$ 
 
-Predictions from the "average" model are centered on the average richness
-observed during training, and the confidence intervals are narrow and
-constant-width.
+Predictions from the "average" model are thus centered on $\mu$, which could
+either be the mean of the raw training richness values, or an output from the
+observer model.  This model's confidence intervals have a constant width that
+depends on the standard deviation of $\epsilon$, which can either be the
+standard deviation of the raw training richness values, or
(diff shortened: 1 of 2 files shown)


sdtaylor:
Observer models are already discussed prior to this, so I see no problem putting those details in this paragraph. Otherwise it's definitely an improvement. It might also help to include the ARIMA(0,1,0) notation for the naive model.

davharris:
Okay, I've added more details and a line in the reviewer response. Might read awkwardly now, but it's probably good to have all this information available. Thanks @sdtaylor for your input above.

sdtaylor:
LGTM

manuscript/draft.Rmd
```diff
@@ -292,18 +292,35 @@ table_data %>%
...
+$\sigma^\mathrm{residual}$ from the observer model; see supplement).
+
+The second baseline, called the "naive" model [@hyndman2014], was a simple
+autoregressive process with a single year of history, i.e. an ARIMA(0,1,0)
+model: 
+
+$$y_t = y_{t-1} + \epsilon_t,$$
+
+where the standard deviation of $\epsilon$ is a free parameter for each site. In
+contrast to the "average" model, whose predictions were based on the average
```
ethanwhite: since we say that the '"naive" model predicts', I think this line should be "whose predictions are based on"

davharris:
Thanks both! Fixed the grammar, and merging now that it's been approved by both of you.

Issue #206, opened 2017-11-17
"Annual" data 


davharris:
What's the best way to clarify the time scale of the data (i.e. routes are visited on a single morning and we use that morning to stand in for the whole year)?

See https://github.com/weecology/bbs-forecasting/pull/205#discussion_r149753730

sdtaylor:
final text in #236

PR #233, opened 2017-12-09
update details of environmental methods

sdtaylor:
These might help with #206. I'll add any changes from there we decide on. 

Saying we did not use the calendar year is an important detail.
Updated months used in the NDVI calcs as per the current code.


davharris:
Good thinking. I agree that this is important.

ethanwhite:
:+1:

PR #234, opened 2017-12-09
Reply intro

sdtaylor:
Other peerj papers have this so I'm assuming we'll have one.


davharris:
This seems like a good start. Do we want to go into more detail, or would that be unnecessary/repetitive? Also, it looks like we aren't given the editor's name (unless I'm just missing it), so maybe we just replace "XXX" with "Editor" in the salutation?

sdtaylor:
I can't see the editors name anywhere either.

sdtaylor:
Of the 4 random rebuttals intros I just read on peerj, this is longer than 3 of them, so I'd say any longer is unnecessary.

ethanwhite:
:+1:

davharris:
In the last sentence, I'd say

>More detailed responses to individual comments are **included** below.

Otherwise, LGTM! 👍

PR #240, opened 2017-12-11
script to archive the forecasts

davharris:
This creates the gzipped csv for zenodo


sdtaylor:
I can't test this cause only you have all the RDS files in the results folder.

PR #241, opened 2017-12-11
track changes for resubmission

davharris:


sdtaylor:
Does latexdiff do image changes? If not it's probably not worth the effort since we just had 1 minor change in a figure. 

Otherwise LGTM

ethanwhite:
Looks like this is missing #239 (Senyondo ref still gives Kironde as Henry's name)?

Also, the diff on the Ye -> Perretti citation isn't working quite right, but there may not be anything we can do about that.

sdtaylor:
@ethanwhite I think because the old Ye citation was taken out it did not get rendered by latexdiff.

davharris:
Yikes, you're right need to regenerate multiple pdfs for #239. Thanks for pointing this out!

Don't think that fixing the other citation is worth the effort right now; the original and final PDFs are both good, it's just the diff that's broken.

I probably could have gotten the images to work by playing with paths and stuff, but it didn't seem worth the effort, given that they're basically unchanged and can be seen by reviewers/editors in a few other places already.

sdtaylor:
so close...

ethanwhite:
Sounds good.

PR #248, opened 2018-01-05
Editor comments

davharris:
These changes are pretty straightforward, I think.  

Thera are a couple of places where I left in the word "have" where the editor suggested omitting it:

>High-profile studies forecasting large biodiversity declines over the coming decades **have** played a large role in shaping the priorities of ecologists and policymakers

and:

>Previous efforts to predict future patterns of terrestrial species richness, and diversity more generally, **have** focused primarily on building species distributions models

We can still remove them if you think going along with the editor would be easier though.

### DIFF
--- a/manuscript/draft.Rmd
+++ b/manuscript/draft.Rmd
@@ -305,8 +305,8 @@ $$y_t = \mu + \epsilon_t.$$
 
 Predictions from the "average" model are thus centered on $\mu$, which could
 either be the mean of the raw training richness values, or an output from the
-observer model.  This model's confidence intervals have a constant width that
-depends on the standard deviation of $\epsilon$, which can either be the
+observer model. The confidence iterval width in this model is constant and a 
+consequence of the standard deviation $\epsilon$, which can either be the
 standard deviation of the raw training richness values, or
 $\sigma^\mathrm{residual}$ from the observer model; see supplement).
 
@@ -332,7 +332,7 @@ function in the package `forecast`; @forecast1] to represent an array of
 different time-series modeling approaches. These models can include an
 autoregressive component (as in the "naive" model, but with the possibility of
 longer-term dependencies in the underlying process), a moving average component
-(where the noise can have serial autocorrelation) and an
+(where the noise may be serially autocorrelation) and an
 integration/differencing component (so that the analysis could be performed on
 sequential differences of the raw data, accommodating more complex patterns
 including trends). The `auto.arima` function chooses whether to include each of
@@ -350,7 +350,7 @@ stacking individual species distribution models; and joint species distribution
 models (JSDMs). Following the standard approach, site-level random effects were
 not included in these models as predictors, meaning that this approach
 implicitly assumes that two sites with identical Bioclim, elevation, and NDVI
-values should have identical richness distributions. As above, we included
+values will have identical richness distributions. As above, we included
 observer effects and the associated uncertainty by running these models 500
 times (once per MCMC sample).
 
@@ -506,7 +506,7 @@ Most models produced confidence intervals that were too narrow, indicating
 overconfident predictions (Figure 5C). The random forest-based SDM stack was the
 most overconfident model, with only `r numbers$rf_coverage_pct`% of observations
 falling inside its 95% confidence intervals. This stacked SDM's narrow
-predictive distribution caused it to have notably higher deviance (Figure 5B)
+predictive distribution caused notably greater deviance (Figure 5B)
 than the next-worst model, even though its point estimates were not unusually
 bad in terms of RMSE (5A). As discussed elsewhere [@harris2015], this
 overconfidence is a product of the assumption in stacked SDMs that errors in the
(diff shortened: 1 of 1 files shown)


davharris:
I've renamed this issue to include all of the editor's suggested changes, since Shawn's suggestions were based on my branch.

manuscript/draft.Rmd
```diff
@@ -75,20 +75,20 @@ ecological models capture the key processes governing natural systems
...
 these models describe individual species' occurrence patterns as functions of
 the environment. Given forecasts for environmental conditions, these models can
-predict where each species will occur in the future. These species-level 
-predictions are then combined ("stacked") to generate forecasts for species 
+predict where each species will occur in the future. These species-level
+predictions are then combined ("stacked") to generate forecasts for species
 richness [e.g. @calabrese2014]. Alternatively, models that directly relate
-spatial patterns of species richness to environment conditions have been
+spatial patterns of species richness to environment conditions are available
 developed and generally perform equivalently to stacked SDMs [@algar2009;
```
ethanwhite: This should either be "have been developed" (the original) or "are available" (i.e., the "developed" needs to be cut as well). I prefer the original but don't care that much.

manuscript/draft.Rmd
```diff
@@ -305,8 +305,8 @@ $$y_t = \mu + \epsilon_t.$$
 
 Predictions from the "average" model are thus centered on $\mu$, which could
 either be the mean of the raw training richness values, or an output from the
-observer model.  This model's confidence intervals have a constant width that
-depends on the standard deviation of $\epsilon$, which can either be the
+observer model. The confidence iterval width in this model is constant and a 
+consequence of the standard deviation $\epsilon$, which can either be the
```
ethanwhite: "iterval" -> "interval", but I'm not sure I love either of these versions. What about:

> In this model the confidence intervals have a constant width that is set by the standard deviation...

manuscript/draft.Rmd
```diff
@@ -332,7 +332,7 @@ function in the package `forecast`; @forecast1] to represent an array of
 different time-series modeling approaches. These models can include an
 autoregressive component (as in the "naive" model, but with the possibility of
 longer-term dependencies in the underlying process), a moving average component
-(where the noise can have serial autocorrelation) and an
+(where the noise may be serially autocorrelation) and an
```
ethanwhite: "serially autocorrelated" (not autocorrelation)

manuscript/draft.Rmd
```diff
@@ -506,7 +506,7 @@ Most models produced confidence intervals that were too narrow, indicating
 overconfident predictions (Figure 5C). The random forest-based SDM stack was the
 most overconfident model, with only `r numbers$rf_coverage_pct`% of observations
 falling inside its 95% confidence intervals. This stacked SDM's narrow
-predictive distribution caused it to have notably higher deviance (Figure 5B)
+predictive distribution caused notably greater deviance (Figure 5B)
```
ethanwhite: I prefer the original here, but if we're going to make the change I think "resulted in notably greater deviance" is better.

I think the right solution here is probably something like this sentence, which address the awkwardness the editor is trying to tackle would be, without making the meaning of the sentence more vague:

> This narrow predictive distribution resulted in the stacked SDM having notably higher deviance...

manuscript/draft.Rmd
```diff
@@ -574,20 +574,20 @@ The most commonly used methods for forecasting future biodiversity, SDMs and
 macroecological models, both produced worse forecasts than time-series models
 and simple baselines. This weakness suggests that predictions about future
 biodiversity change should be viewed with skepticism unless the underlying
-models have been validated temporally, via hindcasting and comparison with
+models are temporally validated, via hindcasting and comparison with
```
ethanwhite: What about taking this a step further to "temporally validated using hindcasting...". I think the editors simplification makes this flow more smoothly.

manuscript/draft.Rmd
```diff
@@ -574,20 +574,20 @@ The most commonly used methods for forecasting future biodiversity, SDMs and
...
 simple baselines. Since site-level richness is relatively stable, spatial
-validation is not enough: a model can have high accuracy across spatial
-gradients without being able to predict changes over time. This gap between
-spatial and temporal accuracy is known to be important for species-level
+validation is not enough: a model be accurate across spatial gradients while
```
ethanwhite: I think you lost a "can" between "model" and "be"

manuscript/draft.Rmd
```diff
@@ -574,20 +574,20 @@ The most commonly used methods for forecasting future biodiversity, SDMs and
...
+being unable to predict changes over time. This gap between
+spatial and temporal accuracy is important for species-level
 predictions [@rapacciuolo2012; @oedekoven2017]; our results indicate that it is
-substantial for higher-level patterns like richness as well. SDMs' poor temporal
-predictions are particularly sobering, as these models have been one of the main
-foundations for estimates of the predicted loss of biodiversity to climate
-change over the past decade or so [@thomas2004; @thuiller2011; @urban2015]. Our
+substantial for higher-level patterns like richness as well. The poor quality 
+temporal predictions of SDM's are particularly sobering because of their importance
+over the last decade or so for predicting biodiversity loss as a response to
```
ethanwhite: Let's also get rid of "or so" here.

ethanwhite (commented):
This looks great overall. Thanks to both of you for getting this taken care of.

I've caught a few typos that were introduced during the editing and also suggested a couple of alternative edits to those made by the editor. It's not uncommon for someone to see a sentence that doesn't work well, try to edit it, and not get it perfect the first time.

I've also noticed a couple of little things that were our original language but that could benefit from minor changes. This is a great time to clean up little language issues since we know that the paper is going in as is.

manuscript/draft.Rmd
sdtaylor: How about "over the last two decades for predicting..." since we have an '04 reference in there.

ethanwhite: Sounds good.

